PR #5, opened 2025-05-29
feat(#18): Ported `UniMap` from Miri into `no_std`

Gitter499:
# TODO

- Evaluate use of `unsafe` in `UniValMap::remove()`

### DIFF
--- a/bsan-rt/src/lib.rs
+++ b/bsan-rt/src/lib.rs
@@ -28,6 +28,7 @@ use libc_print::std_name::*;
 pub use local::*;
 
 mod block;
+mod borrowtracker;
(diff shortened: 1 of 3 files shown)


bsan-rt/src/lib.rs
```diff
@@ -28,6 +28,7 @@ use libc_print::std_name::*;
 pub use local::*;
 
 mod block;
+mod borrowtracker;
```
icmccorm, line 31: Let's change this to `borrow_tracker` to match Miri's implementation—otherwise, everything looks good to go!

icmccorm (approved):

PR #10, opened 2025-06-02
task(#9): Ported `range_map.rs` from Miri

### LINKED ISSUE #9: Port `range_map` to `no_std`
Right now, the `Tree` implementation in Miri uses a custom RangeMap to map a tag, permission, and location. Porting this DS to `bsan-shared/src/range_map.rs`

Gitter499:
## Notes

Miri's implementation uses `Size` (size_t) via rustc_abi. Not sure why since it seems to be treated as u64 anyways. Refactored to using usize, with casts to u64, don't see any
immediate side-effects of this change

## TODO

Running `xb ci` does not run the tests in `range_map.rs`. Modified `bsan-shared/Cargo.toml` to run tests but that did not work. Not sure why tests aren't running or maybe I am not seeing them being run. 

Closes #9


Gitter499:
Rebased into `main`

bsan-shared/src/range_map.rs
```diff
@@ -0,0 +1,334 @@
...
+    v: Vec<Elem<T>>,
+}
+
+impl<T> RangeMap<T> {
+    /// Creates a new `RangeMap` for the given size, and with the given initial value used for
+    /// the entire range.
+    #[inline(always)]
+    pub fn new(size: Size, init: T) -> RangeMap<T> {
+        let size = size.bytes();
+        let v = if size > 0 { vec![Elem { range: 0..size, data: init }] } else { Vec::new() };
```
icmccorm, line 33: Make sure to use methods that take an allocator; we should have this as `RangeMap<T, A>` with `Vec::new_in`. You'll also need to replace `vec!` with an alternate way of creating a new `Vector`, probably another call to `Vec::new_in`.

bsan-shared/src/range_map.rs
```diff
@@ -0,0 +1,334 @@
...
+        slice.iter_mut().map(|elem| (elem.range.clone(), &mut elem.data))
+    }
+
+    /// Remove all adjacent duplicates
+    #[allow(dead_code)]
+    pub fn merge_adjacent_thorough(&mut self)
+    where
+        T: PartialEq,
+    {
+        let clean = Vec::with_capacity(self.v.len());
```
icmccorm, line 232: Should be `Vec::with_capacity_in`

bsan-shared/src/range_map.rs
```diff
@@ -0,0 +1,345 @@
...
+
+#[derive(Clone, Debug)]
+struct Elem<T> {
+    /// The range covered by this element; never empty.
+    range: Range<u64>,
+    /// The data stored for this element.
+    data: T,
+}
+#[derive(Clone, Debug)]
+pub struct RangeMap<T, A: Allocator> {
```
icmccorm, line 26: We can redefine this as `RangeMap<T, A:Allocator = Global>`, where `Global` is imported from the `alloc` crate.

PR #13, opened 2025-06-02
Port a dummy version of `rustc_span::Span` into bsan

### LINKED ISSUE #12: Port `rustc_span` to `bsan-shared`
## Notes

Used by `diagnostics.rs` and `tree.rs`

Gitter499:
## Notes

Adds a `DummySpan` and a `DummySpanData` (which are almost the same thing) into `bsan-rt` for use by `diagnostics` The official Span has a weird macro that allows it to return the correct span type however it was easier to implement a simple `Into` since the `data()` method is only called once in diagnostics and is expected as a `Span`/`DummySpan`

Closes #12


Gitter499:
A `Span` already exists in `bsan-rt/lib.rs` but does not implement  `data()` method or `SpanData` Could move DummySpan and DummySpanData logic into `bsan-rt/lib.rs` or alternatively comment out/delete `Span` as it is not used anywhere.

Gitter499:
## TODO

Rebase all branches that use `DummySpan` and `DummySpanData` (I believe `port_diagnostics` and `port_tree`) to use new `Span` and `SpanData`

PR #15, opened 2025-06-03
Port `diagnostics.rs` from Miri

### LINKED ISSUE #11: Port `diagnostics.rs` to no_std
## Notes

Not necessary but nice to have in-case we have to do some debugging

Gitter499:
## Notes

`diagnostics` and `tree` are somewhat connected so ported some functionality from `tree`. Removed a lot from `diagnostics` as there are a lot of specifics and design considerations made for Miri that may not apply to bsan. In other words, getting proper diagnostics for bsan should be a separate task.

Closes #11


Gitter499:
Rebased onto `port_rustc_span`

PR #25, opened 2025-06-09
Implement Tree Borrows 

### LINKED ISSUE #23: Implement Tree Borrows into `bsan-rt`
## Notes

Assuming most code from Miri's Tree Borrows is ported, must update and track the tree in `__bsan_read`, `__bsan_write`, etc.

## TODO

- [ ] Familiarize with the ported code (it is very well documented) and figure out what methods and techniques are useful, what needs to be added, and what needs to be modified
- [ ] Construct allocation metadata structure
- [ ] Implement `before_memory_access` (https://github.com/rust-lang/miri/blob/master/src/borrow_tracker/tree_borrows/mod.rs#L42-L72)
- [ ] Validate on small test programs (potentially introducing the CI tool here)

Gitter499:
## Notes
- Very rough implementation with a lot of TODOs and spinoff issues.

## TODO
- [x] Fix clippy warnings
- [ ] Rebase against `main`
- [ ] Implement `__bsan_dealloc`

Closes #23


bsan-rt/src/borrow_tracker/mod.rs
```diff
@@ -1,3 +1,178 @@
...
+use crate::diagnostics::AccessCause;
+use crate::span::Span;
+use crate::{AllocInfo, BorTag, BsanAllocHooks, GlobalCtx, Provenance};
+
 #[cfg_attr(not(test), no_std)]
 pub mod tree;
 pub mod unimap;
+
+// The allocator for BorrowTracker
+pub type BtAlloc = BsanAllocHooks;
```
icmccorm, line 15: Let's avoid having a newtype here; we can shorten `BsanAllocHooks` if you think it's too long.

bsan-rt/src/borrow_tracker/mod.rs
```diff
@@ -1,3 +1,178 @@
...
+    global_ctx: &GlobalCtx,
+    // `Some` intializes the tree
+    retag_info: Option<&RetagInfo>,
+) -> Result<(&'static mut Tree<BtAlloc>, &'static Provenance), ()> {
+    // Get global allocator
+    let allocator = global_ctx.hooks().alloc;
+
+    // TODO: Validate provenance and return a "safe" reference
+
+    assert!(unsafe { prov.as_ref().is_some() });
```
icmccorm, line 33: Make sure that these assertions are `debug_assert!`; otherwise they'll be left in release builds.

bsan-rt/src/borrow_tracker/mod.rs
```diff
@@ -1,3 +1,178 @@
...
+// Potential validation middleware should be part of wrapper API?
+
+// TODO: Replace with custom `Result` type
+///
+pub unsafe fn bt_validate_tree(
+    prov: *const Provenance,
+    global_ctx: &GlobalCtx,
+    // `Some` intializes the tree
+    retag_info: Option<&RetagInfo>,
+) -> Result<(&'static mut Tree<BtAlloc>, &'static Provenance), ()> {
```
icmccorm, line 27: These `'static` lifetimes aren't quite right; both the tree and provenance will be heap-allocated, but neither should be live for the duration of the program. I think we should avoid using references until the last second, when we call `tree.perform_access`. That way, we'll have some anonymous `'a` for the call. This would mean moving some of the middleware out of `bt_validate_tree`.

bsan-rt/src/borrow_tracker/mod.rs
```diff
@@ -1,3 +1,178 @@
...
+
+// The allocator for BorrowTracker
+pub type BtAlloc = BsanAllocHooks;
+// TODO: Create trait for this wrapper functionality
+
+// Potential validation middleware should be part of wrapper API?
+
+// TODO: Replace with custom `Result` type
+///
+pub unsafe fn bt_validate_tree(
```
icmccorm, line 22: I definitely think that this is the right approach, since so much is shared between reading and writing. I think that you could take things one step further though, and have another wrapper along the lines of `bsan_access` that calls `tree.perform_access`. Each of `__bsan_read`, `__bsan_write`, `__bsan_retag`, and `__bsan_dealloc` will call `bsan_access`, with the only difference being the type of `AccessKind` passed as a parameter.

PR #26, opened 2025-06-10
fix(#25): Implemented a clean state of Tree Borrows (in-progress)

### LINKED ISSUE #23: Implement Tree Borrows into `bsan-rt`

### LINKED ISSUE #35: Create custom locking mechanism `BsanMutex` for `bsan-rt`

Gitter499:
## Notes
- Very rough implementation with a lot of TODOs and spinoff issues.

## TODO
- [x] Fix clippy warnings
- [x] Rebase against `main`
- [x] Implement `__bsan_dealloc`
    - [x] Handle deallocating the tree from `AllocInfo`/`AllocMetadata`

Closes #23
Closes #35 (maybe)


bsan-rt/src/lib.rs
```diff
@@ -39,14 +47,22 @@ macro_rules! println {
...
+use crate::block::Linkable;
+use crate::borrow_tracker::tree::AllocRange;
 
 pub type MMap = unsafe extern "C" fn(*mut c_void, usize, i32, i32, i32, off_t) -> *mut c_void;
 pub type MUnmap = unsafe extern "C" fn(*mut c_void, usize) -> i32;
 pub type Malloc = unsafe extern "C" fn(usize) -> *mut c_void;
 pub type Free = unsafe extern "C" fn(*mut c_void);
 pub type Exit = unsafe extern "C" fn() -> !;
 
+static ALLOC_COUNTER: core::sync::atomic::AtomicU64 = core::sync::atomic::AtomicU64::new(0);
```
icmccorm, line 64: This should be part of `GlobalCtx` if possible

bsan-rt/src/lib.rs
```diff
@@ -203,32 +220,121 @@ impl Provenance {
...
+    // SAFETY: Locks the `tree_lock` before performing exclusive access via raw *mut pointer
+    unsafe fn lock_tree(&self, tree_lock: &AtomicBool) -> *mut Tree<BsanAllocHooks> {
+        let _tree_guard = TreeGuard::set(tree_lock);
+        unsafe { (&raw mut (*self.0.lock())) as *mut Tree<BsanAllocHooks> }
+    }
+}
+/// The tree pointer is a ~~lazily unitialized~~ global variable stored in a thread safe wrapper with
+/// a mutex
+/// The tree is initialized during an allocation via a call to `bt_init_tree`
+lazy_static! {
```
icmccorm, line 259: The use of `lazy_static!` means that we're allocating one `Tree` to handle every allocation. Instead, we want each `AllocInfo` object to have its own `Tree` instance. Something like:
```
struct AllocInfo {
    ...
    pub tree: Mutex<LazyCell<Tree<BsanAllocHooks>>>>
}
```
This should eliminate the need for the custom `TreeGuard` and `TreeThreadSafe` wrappers.

bsan-rt/src/borrow_tracker/mod.rs
```diff
@@ -1,3 +1,126 @@
...
+#[allow(clippy::result_unit_err)]
+pub fn bt_retag(
+    tree: &mut Tree<BsanAllocHooks>,
+    prov: &mut Provenance,
+    global_ctx: &GlobalCtx,
+    retag_info: &RetagInfo,
+) -> Result<(), ()> {
+    // TODO: Potentially also update prov.alloc_info?
+
+    if tree.is_allocation_of(prov.bor_tag) {
```
icmccorm, line 80: We should wrap the call to `tree.is_allocation_of` in `#[cfg(debug)]`, since it seems to be only used for debugging purposes in the original `Tree` implementation.

bsan-rt/src/lib.rs
```diff
@@ -257,16 +363,83 @@ extern "C" fn __bsan_deinit() {
 /// Creates a new borrow tag for the given provenance object.
 #[unsafe(no_mangle)]
 extern "C" fn __bsan_retag(prov: *mut Provenance, size: usize, perm_kind: u8, protector_kind: u8) {
-    let _ = unsafe { RetagInfo::from_raw(size, perm_kind, protector_kind) };
+    let retag_info = unsafe { RetagInfo::from_raw(size, perm_kind, protector_kind) };
```
icmccorm, line 366: It doesn't seem like we're actually performing the step of creating a new node in the `Tree` here; `bsan_retag` only performs a read access.

bsan-rt/src/borrow_tracker/mod.rs
```diff
@@ -1,3 +1,126 @@
...
+}
+
+// TODO: Replace with custom `Result` type
+/// # Safety
+///
+/// This function should be called as middleware to verify the provenance and optionally
+/// initialize the tree on retags
+#[allow(clippy::result_unit_err)]
+#[allow(private_interfaces)]
+pub unsafe fn bt_validate_prov(
```
icmccorm, line 48: The middleware pattern with `bt_validate_prov` is nice. I think we could change this so that it gets us closer to the tree, though. 

Instead of returning a pointer to AllocInfo, it could return a pointer to the Tree.

```
pub unsafe fn bt_validate_prov(
    prov: *const Provenance,
    ctx: &GlobalCtx,
) -> Result<*mut Tree, ()>;
```

Everything that needs to happen up until the point when we access the `Tree` should happen here. Then, in the calling context, we acquire the lock and proceed with the access.

We also want `bt_validate_prov` to handle a few additional validation steps before we check for an aliasing violation. These include:

* Checking that the allocation ID in `Provenance` provenance matches the one in `AllocInfo`. Terminate with a use-after-free error if not.

* Checking that the pointer's address is within the bounds of the allocation, based on the range information inside `AllocInfo`. Terminate with an access out-of-bounds error if not. 
****

icmccorm (changes requested):
Great work! There are a couple key changes that we still need, but this is close to being ready.

icmccorm:
Closing, since this was merged in #45.

PR #43, opened 2025-06-23
Implement error handling for `borrow_tracker`

### LINKED ISSUE #42: Create an `error.rs` with custom `BsanError` enum and `BsanWrapper` trait (maybe?) to handle `Result`s, program termination, and error handling

Gitter499:
## Notes

Moved around some of the Error types from Tree to make `TreeError`. Added `-Zdylib-lto` flag to `env.rs` which makes build noticeably slower but is required for some procedural macros being utilized in the `thiserror_no_std` crate. 

## TODO

Due to lots of error nesting, it is a little bit tedious to write out Errors. Thinking about refactoring with specialized macros to make error handling less tedious.

Closes #42


bsan-rt/src/borrow_tracker/mod.rs
```diff
@@ -1,3 +1,215 @@
...
+        }
+
+        let mut tree_lock = alloc_info.tree_lock.lock();
+
+        // Cast the Tree void ptr into `Tree` by locking the Tree
+        let tree_ptr = unsafe { &raw mut tree_lock as *mut Tree<BsanAllocHooks> };
+
+        // Dropping the tree lock early as we already have our tree_ptr
+        // TODO: Validate that this drop does not lose us access to the tree pointer
+        core::mem::drop(tree_lock);
```
Gitter499, line 103: Possibly induces a race condition

icmccorm: I think we need to hold onto the lock until after we finish with the `BorrowTracker` object---that way, other threads will be held up within `BorrowTracker::new` until the current execution finishes modifying the `Tree`.

icmccorm: So, instead of manually calling `core::mem::drop`, we'd just have another field within `BorrowTracker`.

bsan-rt/src/borrow_tracker/mod.rs
```diff
@@ -1,3 +1,215 @@
...
 pub mod unimap;
+
+// TODO: Create struct for this wrapper functionality
+
+#[derive(Debug)]
+pub struct BorrowTracker<'a> {
+    prov: &'a Provenance,
+    ctx: &'a GlobalCtx,
+    allocator: BsanAllocHooks,
+    alloc_info: &'a mut AllocInfo,
```
icmccorm, line 31: It's not safe to have this as a mutable borrow; even though we're locking the `Tree`, other threads can still potentially read the `AllocInfo` object.

PR #50, opened 2025-07-08
Added attribution and support for protected tags.

icmccorm:
* Adds support for protectors. 

* Updates our fork of Miri's permissions module to match [2a46692](https://github.com/rust-lang/miri/commit/2a466925c7931bd7c31d6571bca4b32c89aa1af1), which adds a new `Cell` permission type for locations covered by `UnsafeCell`. 

* Updates our Rust fork to [rolling-1.89.0-dev-540ee5b](https://github.com/BorrowSanitizer/rust/releases/tag/rolling-1.89.0-dev-540ee5b), adding support for `Cell` in retags and for specifying the side-effect action type (which is always a read access under Tree Borrows, but could change in a future revision to the model).

* Added some additional notes clarifying that we're using Miri's borrow tracker.


Gitter499 (approved):
Looks great! Nothing to note on my end.

PR #56, opened 2025-08-05
Added revised implementations of `Stack` and `Heap` allocators.

icmccorm:
Originally, our bump allocators for `AllocInfo`  instances had a finite capacity (1024). However, it turns out that "real" programs quickly reach this capacity, even with a free list. This PR refactors everything into two key structs: a `Stack<T>` and a `Heap<T>`. Both are bump allocators that rely on a linked list of page-sized blocks. The `Stack` frees objects by pushing and popping frames. The heap can free an individual object anywhere within any block, adding it to a free list. The `Stack` is single-threaded, while the `Heap` supports concurrent access.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,9 +38,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.19"
+version = "0.6.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


icmccorm:
Credit to [`bumpalo`](https://docs.rs/bumpalo/latest/bumpalo/) and [`bump-scope`](https://crates.io/crates/bump-scope) for inspiration.

PR #62, opened 2025-09-26
Added basic automated benchmarking

Gitter499:
## Features

- Switch between box plots, scatter plot, and time series
- Toggle Miri with logarithmic scaling
- Switch between different architectures and tests

## Features currently in-progress

- Add ASAN
- Automatically pull and run on new programs from `BorrowSanitizer/benches`
- Automatic diff table with different programs on every commit (see https://fast.vlang.io)
- Show benchmark results (stdout) on every bench and highlight common violations in a table
    - Ex. BSAN and ASAN share memory out-of-bounds and use-after-free, so show results for both. BSAN and Miri share     aliasing violations, etc.


icmccorm:
Closing since this is now #127.

PR #63, opened 2025-09-30
Added basic debug logging

mojeanmac:
## Features

- Simple `DebugSummary` struct which implements `Display` for emitting debug logs for the following:
```
__bsan_retag
__bsan_read
__bsan_write
 __bsan_dealloc 
__bsan_alloc 
__bsan_alloc_in_place
```

e.g. running `cargo test --features "debug" --package bsan_rt --lib -- tests::bsan_alloc_increasing_alloc_id --exact --show-output` returns:

```
[DEBUG bsan_rt] [alloc] 0xaaaaf5958b60 @(alloc3, <0>) -> (alloc3, 0xaaaaf5958b60, 64)
[DEBUG bsan_rt] [alloc] 0xffff80001240 @(alloc4, <1>) -> (alloc4, 0xffff80001240, 64)
[DEBUG bsan_rt] [dealloc] 0xffff80001240 @(alloc4, <1>) -> (alloc4, 0xffff80001240, 64)
[DEBUG bsan_rt] [dealloc] 0xaaaaf5958b60 @(alloc3, <0>) -> (alloc3, 0xaaaaf5958b60, 64)
```
- Can be called on operations with the custom `debug_bsan!()` macro
- Should be invoked with `./xb inst --debug program.rs`, theoretically. See TODO

## TODO

- I'm not sure if the `bsan-rt` library is being used anywhere yet, so I haven't had any luck yet in getting the debug statements to appear outside of inline tests. Make sure `println!` in general is still emitted where the library is used.
- ~~Custom panic impl had to be commented out, because `log` uses `std` somehow. Could switch to a more lightweight logging library like https://github.com/knurling-rs/defmt if this breaks stuff~~ switched to `println!`
- Extend to logging `shadow` operations if necessary, + would be nice to toggle them on/off depending on granularity needed
- Extend to `debug_assert` operations if they are used often enough

### DIFF
--- a/bsan-rt/src/lib.rs
+++ b/bsan-rt/src/lib.rs
@@ -84,29 +84,37 @@ struct DebugSummary {
 #[cfg(feature = "debug")]
 impl fmt::Display for DebugSummary {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
-            "[{}] 0x{:x} @({:?}, {:?}) -> ({:?}, {:?}, {:?})",
-            self.op,
-            self.ptr,
-            self.alloc_id,
-            self.bor_tag,
-            self.info.alloc_id,
-            self.info.base_addr,
-            self.info.size
-        )
+        match self.info {
+            AllocInfoSummary::WildCard => write!(
+                f, "[{}] 0x{:x} @(*, {:?}) -> (wildcard)",
+                self.op, self.ptr, self.bor_tag),
+            AllocInfoSummary::Null => write!(
+                f, "[{}] 0x{:x} @({:?}, {:?}) -> (null)",
+                self.op, self.ptr, self.alloc_id, self.bor_tag
+            ),
+            AllocInfoSummary::Valid { alloc_id, base_addr, size } => write!(
+                f, "[{}] 0x{:x} @({:?}, {:?}) -> ({:?}, {:?}, {:?})",
+                self.op, self.ptr, self.alloc_id, self.bor_tag, alloc_id, base_addr, size
+            ),
+        }
     }
 }
 
 #[cfg(feature = "debug")]
 macro_rules! debug_bsan {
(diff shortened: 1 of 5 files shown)


bsan-rt/src/lib.rs
```diff
@@ -330,6 +392,7 @@ unsafe extern "C-unwind" fn __bsan_retag(
     bor_tag: BorTag,
     alloc_info: *mut AllocInfo,
 ) -> BorTag {
+    debug_bsan!("retag", object_addr, alloc_id, bor_tag, unsafe { &*alloc_info });
```
icmccorm, line 395: If we're dealing with a wildcard provenance value (always valid) or a null provenance value (always invalid), then `alloc_info` will be a null pointer, so creating this reference is UB.

bsan-rt/src/lib.rs
```diff
@@ -293,6 +340,21 @@ impl AllocInfo {
...
+
+    #[cfg(feature = "debug")]
+    fn summarize(&self) -> AllocInfoSummary {
+        AllocInfoSummary { alloc_id: self.alloc_id, base_addr: self.base_addr, size: self.size }
+    }
+}
+
+/// A shallow version of `AllocInfo`, for use in debug logging.
+/// Only difference is it does not include the tree_lock `Mutex` field.
+#[cfg(feature = "debug")]
```
icmccorm, line 352: It might be helpful to have this as an enum with `Wildcard` and `Null` variants that don't have any fields.

icmccorm (changes requested):
I tested your changes on top of #64 and found a few things that I had overlooked.

bsan-rt/src/lib.rs
mojeanmac: Yikes! I just pushed a version that uses your suggestion below that displays `-> (wildcard)` or `-> (null)` if the `AllocId` is 0 or 1

icmccorm (approved):

PR #67, opened 2025-10-16
First set of UI Tests

obraunsdorf:
## Porting Miri's tests
- start with `tests/[fail|pass]/[both|tree]_borrows`
	- replace all `//@` directives with `//@run: 1`
	- remove the `retag_data_race_write.rs` test
	- remove all `//~\[[stack|tree].*Error.*` directives
	- remove all `*[stack|tree].stderr` files
	- run `xb ui --bless` ==> **Caveat: I did not verify BSAN's output to detect the correct error. Just that bsan detects an error**
	- deleted cause they need `miri/tests/utils` which depend
		- `fail/tree_borrows/reserved/int-protected-write.rs`
		- `fail/tree_borrows/cell-inside-struct.rs`
		- `fail/tree_borrows/reserved/cell-protected-write.rs`
		- `miri-tests/pass/tree_borrows/cell-alternate-writes.rs`
		- `miri-tests/pass/tree_borrows/cell-inside-struct.rs`
		- `miri-tests/pass/tree_borrows/cell-inside-box.rs`
		- `pass/tree_borrows/formatting.rs`
		- `miri-tests/pass/tree_borrows/end-of-protector.rs`
		- `miri-tests/pass/tree_borrows/reborrow-is-read.rs`
		- `miri-tests/pass/tree_borrows/reserved.rs`
	- BSAN fails to detect errors in (see should-fail directory)
		- protector-write-lazy.rs ==> ???
	- BSAN has false positive in
		- basic_aliasing_model.rs::disjoint_mutable_subborrows() line 163 => string compare?
		- basic_aliasing_model.rs::zst() line 302 ==> raw to reference cast on ZST?
			- maybe unnecessary retag instruction?
		- interior_mutability.rs
	- BSAN returns an internal Error: `Internal(Alloc(StackOverflow))` (see directory `unexpected-internal-error`) for 4 tests that all seem to be related to multi-threading


icmccorm:
> deleted cause they need miri/tests/utils which depend on the miri-extern crate

It would be useful to have these still in-tree so that we can enable them in the future (or, at least, a form that's compatible with our logging.

> BSAN returns an internal error...for 4 tests that all seem to be related to multi-threading

I haven't tested multithreading yet; I'm surprised it worked so well and didn't just segfault.

At least some of the should-fail and false positive errors should be fixed by #68.

PR #82, opened 2025-11-18
Validate TLS across uninstrumented code

icmccorm:
When a pointer is passed to or returned from a function, its provenance is stored in one of two thread-local arrays: `__BSAN_PARAM_TLS` and `__BSAN_RETVAL_TLS`. If instead we modified function signatures to pass and return this information, then that would make our instrumented code incompatible with uninstrumented libraries. MemorySanitizer uses this technique, too.

However, this is also a potential source of false-positive errors. Consider the case where we are in an instrumented function `A` that calls an uninstrumented function `B`. This function receives a pointer, so we pass its provenance through our thread-local array. `B` calls a series of uninstrumented functions, creating a call stack all the way down to the function `Z`, which is instrumented. When we are executing `Z`, we have no way of knowing whether the provenance in our thread-local array is valid for its parameters. Maybe it is, by coincidence. It's common for a series of function calls to have similar signatures. However, we cannot know this for sure.

We can fix this by using frame pointers. Every time we enter an uninstrumented function from an instrumented function, we store a copy of the current frame pointer in a thread-local "marker" variable (`__BSAN_TLS_MARKER`). 
```rust
#[unsafe(no_mangle)]
extern "C" fn __bsan_mark_tls() -> usize {
    unsafe { ptr::replace(&raw mut __BSAN_TLS_MARKER, FramePointer::current().prev()).addr() }
}
```

When enter an instrumented function that might be called from an uninstrumented code (e.g. it has external linkage or is called via a function pointer), we can check to see if our thread-local "marker" matches the frame pointer of our caller (`__bsan_validate_param_tls`). If so, we know our parameter TLS is valid. If not, we wipe it, replacing its contents with wildcard provenance values. We would also change the marker to NULL so that the original caller can know that it crossed into uninstrumented code. When we're back within the body of this caller, we check the marker to see if its been modified and modify the return TLS array accordingly (`__bsan_validate_retval_tls`). 

This adds additional runtime overhead that we expect will be reduced by LTO. If we can guarantee that the entire program will be compiled as a single module, then we can skip TLS validation in all circumstances, except for functions that are called via pointers. This would require some form of alias analysis to identify when a function's pointer is passed to an uninstrumented function.

### DIFF
--- a/bsan-rt/src/local.rs
+++ b/bsan-rt/src/local.rs
@@ -49,7 +49,7 @@ pub static mut __BSAN_PARAM_TLS_MARKER: FramePointer = FramePointer::null();
 /// After validating the parameter TLS, we set this marker equal to the value of
 /// the parameter TLS marker. This will be the null frame pointer if we came from
 /// an uninstrumented function, and it will be the value of our "grandparent" frame
-/// pointer if we came from an instrumented function. The caller of a possibly 
+/// pointer if we came from an instrumented function. The caller of a possibly
 /// uninstrumented function can check this value after the call against the value it
 /// stored in the parameter TLS marker. If they match, then the return value can be
 /// trusted. If not, then we overwrite the return value TLS with wildcard values.


icmccorm:
Okay, so the approach I described above does not work for `aarch64`, or any other architecture where handling tail calls affects the value of the frame pointer.

Rust's global allocator API is implemented as a series of shims, which redirect to either the standard library's implementation or a custom one defined with the [`#[global_allocator]`](https://doc.rust-lang.org/std/alloc/index.html#the-global_allocator-attribute) attribute. For example, the function `__rust_alloc` is compiled as a wrapper, which [performs a tail call](https://github.com/rust-lang/rust/blob/7281a3bc4b89a010aed05528d648450feb8b69d3/compiler/rustc_codegen_llvm/src/allocator.rs#L202) into `__rdl_alloc`. Here's how this shim is emitted:
```rust
// call `__rdl_alloc` within `__rust_alloc`
let ret = bx.call(ty, callee, &args, None);
llvm::LLVMSetTailCall(ret, TRUE);
if output.is_some() {
    bx.ret(ret);
} else {
    bx.ret_void()
}
```
As a result, the frame pointer ends up being offset, so that the value of the frame pointer within `__rust_alloc` is different from the value of the caller's frame pointer within `__rdl_alloc`. This false-positive ends up wiping the TLS within `__bsan_validate_param_tls`. It's not clear to me exactly why this happens; there are other situations I've seen that do not involve these shims. 

We can fix this by moving up one level of indirection. If A calls, B, and B calls a possibly uninstrumented C, then we store A's frame pointer within B before calling C. If C is an alloc shim, its view of B's frame pointer will have been corrupted, but it can still resolve A's frame pointer correctly.

PR #114, opened 2026-01-29
Skip retags for assignment

icmccorm:
Updates to a new version of our 1.95.0 fork, which skips emitting retags for plain assignment statements.


icmccorm:
Note: tests like the following will no longer be UB:
```rust
fn main() {
    let x = &mut 42;
    let xraw = x as *mut _;
    let xref = unsafe { &*xraw };
    let xref_in_mem = Box::new(xref);
    unsafe { *xraw = 42 }; // unfreeze
    let _val = *xref_in_mem;
}
```
Usually, we would retag `_val` after reading it from memory on the last line.

PR #117, opened 2026-02-03
Retrieve source lines for error messages

mojeanmac:
Adds source file lines to diagnostics.
e.g. in miri-tests/fail/both_borrows/alias_through_mutation.rs:
```
help: the accessed tag <TAG>(unprotected) was created here, in the initial state Frozen
      --> bsan/tests/miri-tests/fail/both_borrows/alias_through_mutation.rs:6:14
         |
       6 | *x = &mut *(target as *mut _);
         |

    help: the accessed tag <TAG>(unprotected) later transitioned due to a foreign write
      --> bsan/tests/miri-tests/fail/both_borrows/alias_through_mutation.rs:15:5
         |
      15 | *target = 13;
         |
```

Ports over related lines from the source file for diagnostics with `sanitizer_file::ReadFileToBuffer`. Alternatively, a buffered read could be implemented using `sanitizer_posix::ReadFromFile` that could halt once the desired line is reached.

However, since I'd prefer to (a) minimize low level C++ buffer manipulation and (b) cache the file contents as some diagnostics reference multiple lines from a file, I think this is the best configuration.


icmccorm (commented):
LGTM, other than one comment.

Issue #115, opened 2026-01-30
`Provenance` values are too large for atomics


icmccorm:
Our provenance values are three words:
```rust
#[repr(C)]
pub struct Provenance {
    alloc_id: AllocId,
    bor_tag: BorTag,
    alloc_info: *mut AllocInfo,
}
```
No atomic operations can support values of this size. We need all three components to be loaded in a single atomic operation; otherwise, we'll have data races. 

One option to fix this would be to remove the `AllocId`, and modify the tree so that if a `BorTag` is not present, we treat it as a use-after-free error. This would allow for 128-bit atomic load and store operations, which are relatively well-supported.

Alternatively, we could allocate each node in the tree with a stable pointer, and have provenance just be a single word: a pointer to the node corresponding to that borrow tag. This is similar to what [CapsLock](https://arxiv.org/abs/2507.03344) does.

icmccorm:
Closing, as atomics require independent synchronization regardless of provenance size.

PR #119, opened 2026-02-05
Removes `AllocId` from provenance.

icmccorm:
This begins to address #115 by removing `AllocId` values from provenance. To detect UAF errors, we check if the tree associated with the `AllocInfo` of a provenance value is `None`. We also report a UAF if a tag is missing from a tree, instead of panicking.


mojeanmac (approved):
lgtm!

PR #122, opened 2026-02-12
Add hashbrown `lots_of_insertions` test

kritdass:
This pull request adds the `lots_of_insertions` test from the [`hashbrown`](https://github.com/rust-lang/hashbrown) crate. This test was taking an unreasonable amount of time to run when running `cargo bsan test` on `hashbrown`, so I have duplicated it with a constant parameter allowing one to change the amount of insertions (currently 10). The test stills runs relatively slowly (~30 seconds) but does pass.


icmccorm:
Thanks for the flamegraph! This looks good, just add a link crediting `hashbrown` in the test case and then we'll be good to go.

PR #124, opened 2026-02-23
Print the source location of the error.

icmccorm:
Adds the source location where the error is triggered to our error messages. 

Access out-of-bounds errors have slightly different formatting to better indicate the offset and size of the allocation.


mojeanmac:
lgtm!

PR #129, opened 2026-02-27
Handle allocation in LLVM

icmccorm:
We enable BorrowSanitizer's LLVM pass using [`registerPipelineParsingCallback`](https://llvm.org/doxygen/classllvm_1_1PassBuilder.html#ab9268c557f7572dc641ac05aaff0c2ac). This lets us use it via rustc by passing `-Cpasses=bsan`, as long as we also pass the path to the plugin file using `-Zllvm-plugins`. We can also use plugins with `clang`, if we pass `-fpass-plugin`. However, unlike `rustc` and `opt`, `clang` does not have a way to enable a pass using the parsing callback. Instead, we need to hook our pass into the optimization pipeline using [`registerOptimizerLastEPCallback`](https://llvm.org/doxygen/classllvm_1_1PassBuilder.html#a41abe02ceef34a89fd73bf8e85987bcd). Other sanitizers use this hook, too. 

When we use the optimization callback, our pass ends up running earlier than it would if we used the parsing callback. Switching to the optimization callback ended up breaking two key features, which are fixed in this patch:

* Passes in the late optimization pipeline will be run more than once. We need to ensure that our instrumentation pass is only run once, or else we get false positives.

* We currently use LLVM's `alloc` attribute to identify heap allocation and deallocation functions. Instead of passing provenance to these functions, we just call `__bsan_alloc` or `__bsan_dealloc` at the callsite. If we have a wrapper function that LLVM labels as deallocating its argument, and it calls another function with that label, and the inner function is inlined, then our instrumentation pass will insert two consecutive calls to `__bsan_dealloc`, triggering a false-positive use-after free error. We can avoid this using interception.

This also makes several changes that start shifting operations that are not Tree Borrows-specific into the LLVM component of our runtime.


bsan-pass/BorrowSanitizer.cpp
```diff
@@ -1621,5 +1570,8 @@ bool BorrowSanitizer::instrumentFunction(Function &F,
 
   initializeCallbacks(*F.getParent(), TLI);
   BorrowSanitizerVisitor Visitor(F, *this, TLI, DT);
-  return Visitor.run();
+  Visitor.run();
+
+  F.addFnAttr(Attribute::DisableSanitizerInstrumentation);
```
obraunsdorf: this seems like a safeguard to prevent instrumenting twice as you said in the description is prohibited by this commit. Is this still necessary?

PR #136, opened 2026-03-11
Handle inline assembly conservatively

### LINKED ISSUE #135: Conservatively handle inline assembly
At the moment, we skip instrumenting inline assembly.
```C++
if (CB.isInlineAsm())
    return;
```
Instead, we should clear the provenance of any pointer outputs. We can take instrumentation from what [MemorySanitizer](https://github.com/llvm/llvm-project/blob/3829fdb8afaa4a54d7210a52047e8acb2cf6aa5b/llvm/lib/Transforms/Instrumentation/MemorySanitizer.cpp#L7234) does:
```
    if (CB.isInlineAsm()) {
      // For inline asm (either a call to asm function, or callbr instruction),
      // do the usual thing: check argument shadow and mark all outputs as
      // clean. Note that any side effects of the inline asm that are not
      // immediately visible in its constraints are not handled.
      if (ClHandleAsmConservative)
        visitAsmInstruction(CB);
      else
        visitInstruction(CB);
      return;
    }
```
We will need to add a CLI option to our LLVM pass similar to `ClHandleAsmConservative`, which should be enabled by default. MemorySanitizer does quite a bit of work in `visitAsmInstruction`. For now, we can do something much simpler. We can find each pointer output from the instruction, and set its provenance to a wildcard value, like so:
```C++
setProvenance(/* pointer value */, BS.WildcardProvenance);
```

Gitter499:
### Adds very simple implementation that sets output pointers in inline assembly to have `wildcard` provenance values.

TODOS:
- [x] Figure out the difference between handling it conservatively (default) and without the flag
- [x] Write a simple test case
- [x] Remove `llvm-sparse` artifact from `.devcontainer/Dockerfile` (moved)
- [x] Handle aggregate values

Closes #135

### DIFF
--- a/bsan-pass/BorrowSanitizer.cpp
+++ b/bsan-pass/BorrowSanitizer.cpp
@@ -833,11 +833,10 @@ private:
     }
 
     if (CB.isInlineAsm()) {
-      if (ClHandleAsmConservative) {
+      if (ClHandleAsmConservative)
         visitAsmInstruction(CB);
-      } else
-        // TODO: What is the difference in our case?
-        visitAsmInstruction(CB);
+      else
+        return;
     }
 
     // If we've made it here, then we don't have a hard-coded way to handle this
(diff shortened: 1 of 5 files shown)


icmccorm (commented):
Just one minor change needed to handle provenance of aggregates.

icmccorm (commented):
LGTM after these changes.

PR #147, opened 2026-03-20
Use flow-sensitive provenance stack slot allocation.

icmccorm:
We use a shadow stack to store the provenance values for pointers that are in scope within the current "regular" stack frame. At the moment, we only use this for protected provenance values. When we perform a function-entry retag, we decrement the shadow stack pointer to allocate a slot, where we store the new provenance value. When the function returns, we read the value of the stack pointer to determine how many protectors that we need to remove. 

This is inefficient. Every time we need a new stack slot, we have to load and store the stack pointer. This adds a new method that we can use the layout of the stack ahead of time, as we insert our instrumentation. A single `StackSlotAllocator` keeps track of the offset from the top of the shadow stack frame. When we need a new slot, we increment the offset for the current basic block, and add it to the stack pointer that we loaded in at the beginning of the function (`FrameTop`).
```C++
  Value *allocStackSlot(IRBuilder<> &IRB) {
    Value *SlotOffset = StackSlots.allocSlot(DT, IRB, BS.IntptrTy);
    Value *ProvOffset = IRB.CreateMul(SlotOffset, BS.PL.ProvenanceSize);
    return subtractPointer(IRB, FrameTop, ProvOffset);
  }
```
Offsets are tracked flow-sensitively; the offset at the exit of a function depends on the path taken from the entrypoint. This is necessary since the number of function-entry retags that occur depends on the kinds of arguments being passed to the function, like if the argument is a sum type and only a subset of its variants contain references.


icmccorm:
This also does some refactoring of the way we represent provenance values. Mostly, it removes `WithProvenanceKind` and `ProvenanceKind`, since we can already tell if a provenance value is a scalar of vector value based on its `ElementCount`. 

Helper functions have been moved to the beginning of `BorrowSanitizer.cpp`.

Issue #88, opened 2025-11-28
Thread-safe shadow memory


icmccorm:
Our [`ShadowHeap`](https://github.com/BorrowSanitizer/bsan/blob/main/bsan-rt/src/memory/shadow.rs) isn't thread-safe. Two aspects of this are important: 
1. We do not synchronize accesses to L2 pages.
2. We do not perform atomic shadow accesses for atomic accesses. 
The first should be fixed; the second will be an active research/engineering question. 

If we are writing to shadow memory and an L2 page does not exist, then we create one:
```rust
// src/memory/shadow.rs:
fn ensure_l2(
  &self, 
  hooks: &BsanHooks, 
  idx: TableIndex
) -> AllocResult<NonNull<L2Array<T>>> {
  unsafe {
    let l2_table_ptr: *mut *mut L2Array<T> = &raw mut (*self.table.as_ptr())[idx.l1_index];
    if (*l2_table_ptr).is_null() {
      // allocate a new L2Array<T> using `mmap`
    }
  ...
}
```
We should cover `self.table` in a [`spin::RwLock`](https://docs.rs/spin/latest/spin/rwlock/struct.RwLock.html), which we can lock every time we need to create an L2 page.

icmccorm:
Finished by #172

PR #148, opened 2026-03-27
Synchronizing on Atomic Operations

obraunsdorf:
If the application employs proper synchronization via atomic operations, we want to maintain synchronization guarantees.
Thus we also have to synchronize on shadow heap accesses that are instrumented for `AtomicPtr<T>` operations.
We cannot only make the shadow heap access atomic, instead, we want to make the shadow heap access and application memory access appear as one atomic operation to prevent thread-safety issues like the following

--------------------
- Threads might interleaving the execution in a harmful way

| Thread 1                                       | Thread 2                                      |
| ------------------------------------ | ----------------------------------- |
| `shadow_store_atomic(provenance1)`                               | `shadow_store_atomic(provenance2)`                              |
| `application_store_atomic(ptr1, addr)`               | `application_store_atomic(ptr2, addr)`             |

- Thus execution can be:
  - shadow_store_atomic(provenance1)
  - shadow_store_atomic(provenance2)
  - application_store_atomic(ptr2, addr)
  - application_store_atomic(ptr1,addr)

- leading to desynchronized metadata: `*addr==ptr1, *metadata==provenance2`

--------------------

To solve this, we force synchronization of threads by locking before and unlocking after the two corresponding memory operations (shadow memory and application memory).
As a sweet spot between lock contention and memory space needed for storing the lock guard, we lock the shadow heap on L1 page granularity.

This solves Problem 2. of Issue #88
So #88 can be closed once this is merged.


bsan-pass/BorrowSanitizer.cpp
```diff
@@ -1577,6 +1625,9 @@ void BorrowSanitizer::initializeCallbacks(Module &M,
   BsanFuncDebugPrintDiff = M.getOrInsertFunction(
       kBsanFuncDebugPrintDiff, AL, IRB.getVoidTy(), IntptrTy, PtrTy);
 
+  BsanFuncDebugGetProvenance = M.getOrInsertFunction(
```
icmccorm: We can remove this and other references/declarations/definitions of `__bsan_debug_get_provenance` (aside from `kBsanFuncDebugGetProvenance`), since we're never calling it.

icmccorm:
Closing, since we switched to a fixed shadow mapping. We should revisit these semantics in the future, using additional dedicated shadow regions to contain locks.

Issue #134, opened 2026-03-10
Enable debugging mode with TLS validation.


icmccorm:
Originally, we used frame and instruction pointers to detect when we cross into uninstrumented programs. 

When we call a possibly uninstrumented function, we stored our current frame pointer in a thread-local variable (`__bsan_mark_tls`). Then, when we are within an instrumented function that might be called from an uninstrumented function, we unwind the stack once to check if our caller's frame pointer matches the "marker" we stored in the thread-local variable (`__bsan_validate_param_tls`). If so, then we set the value of this marker to null. When we return, the caller will see that the marker is null (`__bsan_validate_retval_tls`), and it will know that it can trust the provenance for the return value. If the frame pointer and the marker do not match, then we assign wildcard provenance values to all parameters, and we leave the marker as-is. The caller will see that the marker has not changed, and will know not to trust the return values' provenance.

This approach does not always work, and it is prone to false positive and negative errors. We will no longer use it in #129. However, it is still useful for debugging purposes. We should add an option to re-enable it in our LLVM pass for future use. This will require adding a new CLI option to our LLVM pass, as well as a way to enable debug logging in the LLVM component of our runtime.

icmccorm:
Fixed in #144

PR #164, opened 2026-04-21
Use the shadow stack for everything.

icmccorm:
Closes #153 and #134. Much like MemorySanitizer and DataflowSanitizer, BorrowSanitizer uses two thread-local arrays to store the shadow values of parameters and return values (`__BSAN_PARAM_TLS`, `__BSAN_RETVAL_TLS`).

This switches us over to using a single "shadow stack" to store these values (`__BSAN_PROV_STACK`) as well as every other provenance value associated with a pointer that's loaded into a register. All retags and pointer loads store provenance to the shadow stack.

There are three distinct regions within a function's shadow stack:
1. Parameter provenance
2. Function entry retags
3. Loads, return values, and regular retags. 

We track the offset for region 2 separately from region 3 so that we can remove the protectors created by function-entry retags before the function returns. All values on the shadow stack are contiguous - there are no gaps where stale provenance values could be read.


mojeanmac (approved):
LGTM!

Issue #165, opened 2026-04-25
Garbage Collection Prerequisites


icmccorm:
There are a few tasks that we need to finish as a prerequisite for implementing a deferred reference-counting garbage collector. 

* When our runtime starts, before we initialize interceptors but *after* we initialize our global state (`__bsan_init`), we want to start a new thread, which will be our garbage collector. For now, this thread can do nothing, but it should exit gracefully when the sanitizer runtime dies. 

* Every time a new thread is created, it should register its state in a `LocalCtx` object in our runtime. The `GlobalCtx` should maintain a `RwLock<HashMap<ThreadId, LocalCtx>>`, with `ThreadId` being an atomic counter similar to `AllocId` and `BorTag`. The `LocalCtx` should contain a pointer to the shadow stack, the length of the shadow stack allocation, and a pointer to where the thread's `__BSAN_PROV_STACK` is stored. A prior PR (#80) had a version of this, but it needs to be adapted and updated to where we are now. 

Each of these can be a separate PR.

PR #167, opened 2026-04-29
Add GC thread

mojeanmac:
This pr addresses the first part of #165, creating a new gc thread at initialization that does nothing until bsan deinits.

It uses the `sanitizer_common` wrappers for OS-level threads for starting and stopping (`internal_start_thread` / `internal_join_thread`) and atomic locking for the stop flag.


icmccorm:
LGTM!

PR #168, opened 2026-05-02
Remove separate declarations file.

icmccorm:
Most LLVM sanitizers keep everything in a single *.cpp file. As preparation for merging the LLVM components upstream, we should gradually integrate each of the *.cpp files into a single `BorrowSanitizer.cpp`.

However, certain components are large enough (e.g. the `ShadowStackAllocator`) that they should live in a separate file. AddressSanitizer does this already, with [`ASanStackFrameLayout.cpp`](https://llvm.org/doxygen/ASanStackFrameLayout_8cpp_source.html). That's not necessary for this change, but it may appear in future PRs.


mojeanmac (approved):
LGTM! I like the naming conventions better than `kBsanFuncName`

PR #169, opened 2026-05-04
Add a nop mode for testing the LLVM wrapper.

icmccorm:
Setting `BSAN_NOP` will now configure `cargo bsan` to use just the LLVM runtime component, without the Rust core. This will be useful for testing the wrapper and isolating bugs to specific components of the runtime.


mojeanmac (approved):
LGTM, this should make testing it on real world crates doable now

PR #170, opened 2026-05-04
Init/deinit local ctx

### LINKED ISSUE #165: Garbage Collection Prerequisites

mojeanmac:
This PR adds back in a modified `local.rs` to the runtime to initialize and deinitialize local contexts, from `BsanThread::Init` in the runtime which calls `lib::__bsan_local_init` which calls `local::init_local_ctx` and finally `global::register_thread`. We deinit by following the same path through modules starting with `BsanThread::Destroy`.

We also add a new `ThreadId` type to distinguish `LocalCtx`s in the `GlobalCtx`, which is initialized with a global counter similar to `AllocId` and `BorTag`.

So far, `stack_top` and `stack_ptr` are unused:
```rust
pub struct LocalCtx {
    thread_id: ThreadId,
    stack_top: NonNull<Provenance>,
    stack_ptr: *mut NonNull<Provenance>,
}
```

Closes #165

### DIFF
--- a/bsan-rt/llvm-wrapper/bsan.cpp
+++ b/bsan-rt/llvm-wrapper/bsan.cpp
@@ -249,6 +249,10 @@ SANITIZER_WEAK_ATTRIBUTE void __bsan_internal_init() {}
 
 SANITIZER_WEAK_ATTRIBUTE void __bsan_internal_deinit() {}
 
+SANITIZER_WEAK_ATTRIBUTE void __bsan_local_init(Provenance **prov) {}
+
+SANITIZER_WEAK_ATTRIBUTE void __bsan_local_deinit() {}
+
 // Weak tagging operations
 SANITIZER_WEAK_ATTRIBUTE BorTag __bsan_new_bor_tag() { return 0; }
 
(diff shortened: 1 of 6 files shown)


icmccorm:
LGTM!

PR #171, opened 2026-05-04
Remove unused/unneeded `Provenance` helpers.

icmccorm:
This removes several helper classes for supporting scalable vectors of provenance values. This includes:
* `ProvenanceScalar` and `ProvenanceVector`
* `ProvenancePtr` and all of its subclasses
* `DynSize`

It also relocates `ProvenanceDesc` and related functions into the separate `Provenance.cpp` file. If in the future BorrowSanitizer needs to support scalable vectors, these might need to be added back in some form. However, at that point, our provenance metadata will likely have a different form, so these functions will need to be reimplemented anyway. At the moment, these components just tend to make development trickier.


mojeanmac (approved):
LGTM!

Issue #72, opened 2025-11-03
Switch to using a preallocated shadow memory space.


icmccorm:
At the moment, we use a shadow memory scheme similar to Valgrind, where we allocate pages on-demand using a two-level page table. This is expensive. Every time we store a provenance value to our shadow memory, we need to ensure that we've mapped a page for the corresponding chunk of the address space:
```rust
let idx = TableIndex::new(addr);
unsafe {
  let l2_page = self.ensure_l2(hooks, idx)?;
  ...
} 
```
Instead, it would be great if shadow memory was preallocated, so that accessing shadow memory would only require bit shifting the pointer to its shadow address. 

At the moment, LLVM's [TypeSanitizer (TySan)](https://clang.llvm.org/docs/TypeSanitizer.html) uses an 8:1 ratio of shadow memory to main memory. All we need is a 3:1 ratio (each pointer has three words of memory for the `AllocId`, `BorTag`, and `AllocInfo` address), so it should be feasible to use their approach.

TySan computes how many bits are significant in a pointer, and uses this to determine [the correct mapping](https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/tysan/tysan_platform.h) from shadow to main memory. Then, during initialization of the runtime, it [maps a massive chunk of memory](https://github.com/llvm/llvm-project/blob/fa2c5fe21df42accfcd4498aa7e6b2ae943357ea/compiler-rt/lib/tysan/tysan.cpp#L500), covering the entire shadow space:
```cpp
if (!MmapFixedNoReserve(ShadowAddr(), AppAddr() - ShadowAddr()))
  Die();
```
Shadow addresses are stored as global variables, which are loaded by [the LLVM pass](https://github.com/llvm/llvm-project/blob/fa2c5fe21df42accfcd4498aa7e6b2ae943357ea/llvm/lib/Transforms/Instrumentation/TypeSanitizer.cpp#L491).
```cpp
Instruction *TypeSanitizer::getShadowBase(Function &F) {
  IRBuilder<> IRB(&F.front().front());
  Constant *GlobalShadowAddress =
      F.getParent()->getOrInsertGlobal(kTysanShadowMemoryAddress, IntptrTy);
  return IRB.CreateLoad(IntptrTy, GlobalShadowAddress, "shadow.base");
}
```
[...]

PR #172, opened 2026-05-04
Adjust shadow memory APIs to support refcounting and fixed allocation space.

icmccorm:
This PR makes a few changes to our shadow heap that will simplify switching over to a pre-allocated layout (#72). The function `__bsan_shadow` returns a pointer to the shadow heap location for a given address. Both read and write accesses will allocate a page within the shadow heap. I do not expect this to have a significant impact on performance, under the assumption that most shadow memory pages are both read and written.

We also have `__bsan_rc_store`, which looks a lot like `__bsan_shadow_store` does now. It receives a provenance value and a pointer to a shadow memory location. It does not calculate a shadow address; it receives one from a separate call to `__bsan_shadow`. At the moment, it just stores the provenance value to the location. However, in the future, it will also need to increment and decrement the reference counts of the provenance being stored, and the provenance being replaced.


mojeanmac:
LGTM! I pulled and made the same fix to the merge, but slower

icmccorm:
Oh whups, sorry!

PR #176, opened 2026-05-06
Instrument variadic arguments

Gitter499:
## TODO

This PR does not handle consuming the variadic arguments and does not update `VaList<'>` We need to read variadics from uninstrumented frame contexts. This will be done in a subsequent PR.

Closes #154


icmccorm:
The current version looks great to me. However, I thought of a potential issue surrounding how variadics are passed that we need to tackle. If we are being called from an uninstrumented context, then we cannot rely on `VarArgCounter` having the correct value. We want `__bsan_validate_params` to reset the counter to zero. We can load the value of the counter after this function gets called within `initStack`. 

This also does not handle reading values out of the variadic array yet. We can leave that for a subsequent PR, or make it part of this one.

PR #177, opened 2026-05-08
Update fork of Rust to add taint markers

icmccorm:
This updates our Rust fork to [rolling-1.97.0-dev-1e214e7](https://github.com/BorrowSanitizer/rust/releases/tag/rolling-1.97.0-dev-1e214e7). It removes the component `rustc-dev`, which is no longer necessary. We used this when we had the `bsan-driver` compiler plugin.

This change adds a new flag `-Zcodgen-emit-taint`, which will emit function calls to mark when pointers and places become exposed to certain unsafe operations. 

When a reference is cast into a pointer, or vice versa, we "mark" it with the function `ptr __rust_taint_reg(ptr)`, creating a new alias. For example:
```Rust
let x = &0;
let y = x as *const _;
```
The code above would compile to:
```LLVM
%x = ...
&y = __rust_taint_reg(%x)
```
The same thing happens to pointers within a value that is transmuted. We treat `mem::transmute` as entirely opaque, except in cases where it becomes a nop, which we skip. When the type of a place is transmuted, we mark its contents as tainted using `void __rust_taint_mem(ptr)`. Any pointer that is loaded from a tainted place becomes tainted. This adds a stub implementation of a taint analysis pass.

### DIFF
--- a/bsan-pass/CMakeLists.txt
+++ b/bsan-pass/CMakeLists.txt
@@ -13,12 +13,14 @@ set(BSAN_SOURCES
   BorrowSanitizer.cpp
   BorrowSanitizerPass.cpp
   Provenance.cpp
+  TaintAnalysis.cpp
 )
 
 set(BSAN_HEADERS
   BorrowSanitizer.h
   BorrowSanitizerPass.h
   Provenance.h
+  TaintAnalysis.h
 )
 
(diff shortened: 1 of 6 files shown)


mojeanmac (approved):
LGTM!

PR #178, opened 2026-05-11
Fixed nop retags by splitting up our runtime components.

icmccorm:
Our nop mode was not actually nop-ing. There were a few functions and statics (`__bsan_shadow`, `__bsan_rc_store`, `__BSAN_HAD_ERROR`, `__BSAN_BOR_TAG_CTR`) that did not have corresponding weak definitions in the LLVM wrapper.

This PR adds these definitions and fixes another, related problem. Originally, we combined the LLVM and the Rust runtimes into a single static archive, but this prevents strong symbols from overriding weak symbols. Up until now, we were able to accidentally avoid this problem due to the missing weak symbols. 

This PR solves this by linking our LLVM and Rust runtimes separately, with the Rust one coming first. Our `cargo-bsan` plugin uses two separate environment variables (`BSAN_RT_LLVM`, `BSAN_RT_RUST`) to record the paths to each static archive throughout the build process. This also removes the `BsanRtCore` step from our build script. When using `xb`, the keyword `bsan-rt` now only refers to the Rust component, while `compiler-rt` refers to the LLVM wrapper.

### DIFF
--- a/bsan-rt/Cargo.toml
+++ b/bsan-rt/Cargo.toml
@@ -14,7 +14,7 @@ smallvec = "1.15.1"
 bitflags = "2.11.1"
 
 [lib]
-name = "bsan_rt_core"
+name = "bsan_rt"
 crate-type = ["staticlib"]
 
 [features]
--- a/bsan-rt/llvm-wrapper/bsan.cpp
+++ b/bsan-rt/llvm-wrapper/bsan.cpp
@@ -30,6 +30,14 @@ THREADLOCAL void *__bsan_marker = nullptr;
 SANITIZER_INTERFACE_ATTRIBUTE
 THREADLOCAL Provenance *__bsan_shadow_stack = nullptr;
 
+// A flag set by the Rust core runtime to indicate to the LLVM
+// wrapper that an error has occurred.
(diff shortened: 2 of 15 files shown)


icmccorm:
A couple of notes:
* Instead of setting `BSAN_NOP`, you use the `--nop` flag.
* I added a test for this in `test-cargo-bsan`. There's now a unit test with UB, and it's only enabled for `--nop`.

icmccorm:
@mojeanmac this should also make rebuilds less annoying. Instead of `cargo clean`-ing the entire `target` folder before `test-cargo-bsan`, this will only remove the instrumented binary, if it exists. We will not longer need to rebuild our toolchain every time we execute `./xb ui`. We'll still need to delete and rebuild the sysroot if the instrumentation pass changes, though.

mojeanmac (approved):
LGTM! Does`--nop` also work with `xb inst` or would you have to set the environment flag?

icmccorm:
> LGTM! Does`--nop` also work with `xb inst` or would you have to set the environment flag?

Oh whups, yeah, we'll need a separate `--nop` flag for `./xb inst`

PR #180, opened 2026-05-12
Remove images that are more than one month old

icmccorm:
Adds a final `cleanup` step to building our Docker image, which removes any images that are more than one month old.


mojeanmac:
LGTM!

PR #181, opened 2026-05-12
Fixes `musttail`.

icmccorm:
Fixes one of Miri's test cases in `should-fail` that was related to tail calls.

When a tail call happens, we need to pop the current frame, deallocate all `allocas`, and then store the arguments to the call at the top of the frame header, as if the current function returned and then our caller immediately called the function with the `musttail` annotation.


icmccorm:
Closing in favor of #183.

PR #182, opened 2026-05-12
Update Tree Borrows

mojeanmac:
This PR updates Tree Borrows to the latest Miri model ([072a9fa](https://github.com/rust-lang/miri/commit/072a9fab2041b21cce8d4f21aa84ec9ce7657414)) some files are specifically based on slightly older commits, but you can find them in the top comment of each module.

Some major deviations from Miri:
- We don't use Stacked Borrows, so we don't have a `borrow_tracker` parent folder. Instead, we have our top level functions (`alloc`, `dealloc`, `retag`, etc) live in `borrow_tracker.rs`. This also means we don't need to use Miri's `tb_*` functions in `tree_borrows/mod.rs`.
- We're holding off on adding implicit writes for now, even though the latest commit has them.

What does this change?

- The original `borrow_tracker` folder has been replaced with `tree_borrows` plus `borrow_tracker.rs`
- New data structures (`DedupRangeMap` replaces `RangeMap`) 
- Support for wildcard provenance on the Rust side (cpp side TODO)
- `Active` allocations are now `Unique`, to follow the semantics of the Tree Borrows PLDI '25 paper.

### TODO:

Tests (immediate)

- [x] Fix `box-cell-alias.rs` (should fail)
- [x] Fix `thread_park_isolated.rs` (should pass)
- [x] Investigate why `interior_mutability.rs` triggers `FnExit::Display`
- [x] Investigate loss of error msg granularity in `error_range.rs` and `dealloc_against_protector.rs` (intended behavior)

Other

- [x] Check which broken tests now pass! (Will be covered by #183)
- [x] Remove unused functions
- [ ] Implement lazy tree allocation (or have a student do it?)


bsan-rt/src/borrow_tracker.rs
```diff
@@ -212,40 +193,37 @@ impl<'b> BorrowTracker<'b> {
...
                     alloc_id,
                     span,
-                    alloc::alloc::Global,
                 )?;
             }
         }
 
         // base offset should be the offset, from zero, where the retag is taking place within the allocation.
-        let child_params = ChildParams {
+        let _ = self.tree_mut().new_child(
```
icmccorm: Can we have `new_child` return a `Result` to avoid the `let _`?

config.toml
```diff
@@ -1,8 +1,8 @@
 artifact_url = "https://github.com/BorrowSanitizer/rust/releases/download/"
-tag = "rolling-1.97.0-dev-1e214e7"
-sha = "1e214e7986bf2fa30f080d6931cb67538665a624"
+tag = "rolling-1.97.0-dev-f200f92"
```
icmccorm: We want to use the latest version of the fork, right? Or was there an error?

bsan-rt/src/tree_borrows/perms.rs
```diff
@@ -1,97 +1,57 @@
...
     /// This is required for LLVM IR pointers that are `dereferenceable` (and also allows `noalias`).
-    StrongProtector = 2,
+    StrongProtector,
 }
 
-/// The activation states of a pointer.
 #[repr(u8)]
+#[allow(unused)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum RetagPtrKind {
```
icmccorm: We don't need this anymore, it's replaced by `RetagFlags`.

icmccorm (approved):
LGTM - with a few changes.

icmccorm:
>  Investigate loss of error msg granularity in error_range.rs and dealloc_against_protector.rs

This seems to be intended behavior - Miri's [latest version](https://github.com/rust-lang/miri/blob/48b0b923635281fad0c7597c7c704a7d7fe10cf2/tests/fail/tree_borrows/error-range.stderr#L2) of the error message for this test appears to match the output from this PR.

config.toml
mojeanmac: oops- my mistake!

PR #183, opened 2026-05-13
Update our clone of Miri's test suite.

icmccorm:
This PR finishes adding support for each of Miri's passing tests (from our filtered subset) with the exception of variadics, which will be handled by #176. 

It fixes the following bugs:
* `musttail` calls should be treated as equivalent to a `ret`. 
* We should not call `__bsan_internal_deinit`. Instead, we need to leak our resources for the kernel to reclaim. Otherwise, if the main thread exits before any child threads do, then we will unmap our shadow heap, leading to a use-after-free. This is the heisenbug that was plaguing us in previous PRs with `scope.rs`. Alternatively, we could make it the responsibility of the last exiting thread to deinitialize our runtime.
* Temporarily removes our interceptor for `__cxa_thread_atexit_impl`, which incorrectly used `cxa_atexit`. We need separate handling for this.


icmccorm:
@mojeanmac the tests you moved over in [8fe5726](https://github.com/BorrowSanitizer/bsan/pull/183/commits/8fe572673370f0a3f9634182a6569d6c44d93a6d) are failing, but not for the right reason. 
```
let fault_addr = unsafe { (*info).si_addr().addr() };
```
This looks like a false positive.

PR #127, opened 2026-02-25
Port benchmarking pipeline and visualization

Gitter499:
Ported over benchmarking to the new code

TODO

- [x] Integrate `bencher`

PR #184, opened 2026-05-13
Add a new benchmarking pipeline using `github-action-benchmark`

icmccorm:
Adds support for using [`github-actions-benchmark`](https://github.com/benchmark-action/github-action-benchmark) to generate a dashboard showing median relative execution time. Results are stored in a dedicated `gh-pages` branch, which includes data from previous runs.

The [`index.html`](https://github.com/BorrowSanitizer/bsan/blob/gh-pages/index.html) file generated by this tool can be modified and preserved in subsequent runs, which will let us keep much of @Gitter499's work in #127, while relying on the CI infrastructure from this preexisting action.


Gitter499:
Woah! This looks great. I just had a couple questions regarding the implementation. It looks like we are using cron scheduling to trigger the GitHub action. I was wondering how `github-action-benchmark` will display the graph in this case. From their documentation, it seems like they trigger every commit and use the commit hash to display the x-axis. Since we're using cron, would the x-axis be the `head` commits and update every time the action is ran? Is there anyway to see the charts locally or would we have to wait for this PR to get merged in order for the action to run.

Also, it seems like `github-action-benchmark` uses the same way of transmitting the benchmark data over to the frontend through `window.BENCHMARK_DATA` as we did with the old benchmarking code, so it is possible to write a custom frontend.

icmccorm:
@Gitter499 you can checkout what the results look like [here](https://borrowsanitizer.github.io/bsan/).

> I was wondering how github-action-benchmark will display the graph in this case. From their documentation, it seems like they trigger every commit and use the commit hash to display the x-axis. Since we're using cron, would the x-axis be the head commits and update every time the action is ran?

Ahh, gotcha, yeah, we should have a different trigger than `cron`. I'll update this to run for every new commit to `main`

icmccorm:
> Is there anyway to see the charts locally or would we have to wait for this PR to get merged in order for the action to run.

You can view the charts locally by checking out the `gh-pages` branch, and then manually updating the .js file with new data from running the data collection script. But yeah, it doesn't seem like there's a great way to just automatically view the latest charts without a PR. 

We can add a bench run to our CI for PRs though, as a way to compare the performance of WIP changes against the latest results on `main` The action has additional support for adding alerts for performance regressions, but I haven't looked into that yet.

PR #187, opened 2026-05-14
Get repo stats and check for fixed tests

mojeanmac:
Using `xb stats` you can now see how many true/false positive/negative tests we have. 

The terminology is "positive for bugs" vs "negative for bugs".

The current snapshot looks like:
```
True negative (pass) tests: 282 (99.30%)
  ├─ original tests: 10
  └─ miri tests: 272
True positive (fail) tests: 72 (59.50%)
  ├─ original tests: 8
  ├─ mirilli tests: 0
  └─ miri tests: 64
False positive (should pass) tests: 2 (0.70%)
False negative (should fail) tests: 49 (40.50%)
  ├─ aliasing: 32
  │  ├─ both_borrows: 4
  │  ├─ stacked_borrows: 1
  │  └─ tree_borrows: 27
  │     └─ wildcard: 27
  └─ baseline: 17
     ├─ dangling_pointers: 3
     ├─ intrinsics: 2
     ├─ provenance: 7
     └─ tls: 1
```

`xb fix` runs all tests in the `should-*` categories with a 10s timeout. The current snapshout outputs:

```

SHOULD-PASS: 0/2 tests correctly passed with no errors (0 aborted after 5s).

SHOULD-FAIL: 2/49 tests failed with bsan errors (3 aborted after 5s).

  ✗ Test failed with output: tests/miri-tests/should-fail/baseline/zst_local_oob.rs
    error: Undefined Behavior: an access of size 128b at offset 0xffffffffb1a02331 is out of bounds for alloc277 of size 1b.
       --> /root/.rustup/toolchains/bsan/lib/rustlib/src/rust/library/std/src/sys/pal/unix/stack_overflow.rs:107:35
        |
    107 | let fault_addr = unsafe { (*info).si_addr().addr() };
        |
    
    stack backtrace:

  ✗ Test failed with output: tests/miri-tests/should-fail/baseline/reading_half_a_pointer.rs
    error: Undefined Behavior: an access of size 128b at offset 0xffffffffa3466240 is out of bounds for alloc282 of size 24b.
       --> /root/.rustup/toolchains/bsan/lib/rustlib/src/rust/library/std/src/sys/pal/unix/stack_overflow.rs:107:35
        |
    107 | let fault_addr = unsafe { (*info).si_addr().addr() };
        |
    
    stack backtrace:
```

### TODO:
- [x] Add this summary to the end of `xb ui`
- [x] Add another command to test all the `should-*` cases
- [x] fix thread termination

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -124,6 +124,7 @@ name = "bsan"
 version = "0.1.0"
 dependencies = [
  "colored",
+ "libc",
  "regex",
  "rustc_version",
  "tempfile",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ ui_test = "0.30.2"
 rustc_version = "0.4"
 regex = "1.12.2"
 tempfile = "3"
+libc = "0.2.186"
 
 [[test]]
 name = "ui"
--- a/bsan-script/src/commands.rs
+++ b/bsan-script/src/commands.rs
@@ -57,6 +57,7 @@ impl Command {
                 Ok(())
             }),
             Command::Inst { file, debug, args } => Self::inst(env, file, debug, &args),
+            Command::Fix => Self::fix(env),
             Command::Stats => Self::stats(env),
         }
     }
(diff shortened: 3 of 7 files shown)


joshsunshine:
I think I mostly understand this report, but what are "bsan tests" vs "miri tests"? I assume "true negative (pass) tests" tests that correctly evaluate a program without encountering an error. Is that right? What would a "bsan test" be in that category?

icmccorm:
> I assume "true negative (pass) tests" tests that correctly evaluate a program without encountering an error. Is that right? What would a "bsan test" be in that category?

The "bsan test" category includes tests that we write ~~in addition to the ones we import from Miri~~. These tests are categorized in the same way as the "miri tests". None of these tests are false positives or negatives, since they were added to model correct behavior after we had supported a particular feature.

joshsunshine:
> The "bsan test" category includes tests that we write in addition to the ones we import from Miri. 

@icmccorm Sorry to be pedantic, but "in addition" is confusing me in this sentence. I think you mean that the "bsan test" category includes only bsan- team-written tests. The "miri test" category is for tests written by the miri team. Is that correct?

icmccorm:
Yes, that's correct!

icmccorm:
@mojeanmac It would be helpful to have this show a percentage.

PR #189, opened 2026-05-15
Avoid treating wildcard provenance as a nop.

icmccorm:
If a provenance value is entirely zeroed-out, we treat it as omnivalid. All memory accesses that use it as validation will succeed. We have been referring to this as having "wildcard" semantics, but that is only partially true. We want to support actual "angelic" wildcard provenance, where we attempt to resolve accesses using pointers with wildcard provenance to allocations that have been "exposed" via pointer to integer conversion. 

This PR skips checks for pointers that do not have any provenance at all, while retaining checks for pointers that have any provenance value, even if it's wildcard. This allows us to differentiate between pointers that we do not support instrumenting, which will not be checked, and those that we want to instrument using wildcard semantics.

### DIFF
--- a/bsan-pass/BorrowSanitizer.cpp
+++ b/bsan-pass/BorrowSanitizer.cpp
@@ -651,27 +651,27 @@ private:
 
   void instrumentRetagReg(CallBase &CB) {
     IRBuilder<> IRB(&CB);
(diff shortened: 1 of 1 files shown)


mojeanmac (approved):
LGTM

PR #191, opened 2026-05-18
Add UI tests from MiriLLI

icmccorm:
This adds a UI test for a bug that MiriLLI found in `flate2`. This was originally reported in [#392](https://github.com/rust-lang/flate2-rs/issues/392) and fixed in [#394](https://github.com/rust-lang/flate2-rs/pull/394).

Tests that replicate MiriLLI results will be placed in `./tests/fail-dep/mirilli`.

The original test (`deflate::tests::drop_writes`) hangs in our current version, but the bug can be found in multiple test cases. The unit test being added in this PR is `deflate::tests::zero_length_read_with_data`.


mojeanmac:
LGTM! Compatible with `xb stats` :)

PR #193, opened 2026-05-23
Avoid loading and storing the shadow stack pointer.

icmccorm:
This PR applies several edge-case optimizations to avoid having to load and store the shadow stack pointer. It also shifts the responsibility of handling shadow stack updates from `BorrowSanitizerVisitor` to `ShadowStackAllocator`, which leads to a slightly cleaner implementation.


mojeanmac:
LGTM. I think this breaks variadics though (at least my merge does) or they were already broken on main.

PR #202, opened 2026-05-28
Moved tests related to performance into should-pass

icmccorm:
One folder of tests in `excluded` contained tests that we do want to support, but that hang  when backtraces are enabled. We should count these in our total for supported test cases.


mojeanmac:
LGTM! Should probably make an issue to fix the backtrace bug. This checks out
```
False positive (should pass) tests: 10 (3.42%)
  └─ performance: 8
     ├─ backtrace: 2
     └─ panic: 5
```

PR #203, opened 2026-05-28
Add dedicated "omnivalid" provenance.

icmccorm:
This PR introduces a new kind of provenance value: omnivalid. We now have four kinds of provenance values:
* *Omnivalid* - `{0, null}` - Any access using a pointer with this provenance value is ignored. These values are assigned to allocations that we can skip checking, or that we do not support instrumenting.
* *Null* - `{1, null}` - Any access using a pointer with this provenance value is undefined behavior.
* *Wildcard* - `{2, null}` - Any access using a pointer with this provenance value is optimistically valid, depending on which provenance values have been "exposed" via integer to pointer conversion.
* *Concrete* - `{>2, nonnull}` - A specific access permission associated with an allocation.

This also temporarily removes our taint analysis, which will be back in a subsequent PR for which this is a prerequisite.


mojeanmac (approved):
LGTM! ATM it isn't obvious to me how this will make adding wildcard support easier. Is it that wildcard will be a less permissive version of omnivalid when exposed provenances are involved?

icmccorm:
> Is it that wildcard will be a less permissive version of omnivalid when exposed provenances are involved?

Yep! If we have omnivalid provenance, then we can ignore a memory access entirely. If we have wildcard provenance, then we have to validate that access using Tree Borrows' optimistic semantics.

PR #207, opened 2026-06-01
Add custom allocator backends with profiling

### LINKED ISSUE #201: Support multiple allocator backends for different allocation strategies
## Goal

Get a holistic understanding of allocation performance in the current implementation of Tree Borrows 
 -> Guide allocator design

## Insights

Added different allocator backends as our global allocator (mimalloc, via static compile flag directives (enabled by `--features` when building). In order to prevent kernel panics, we must provide a free list such that we do not return pages to the OS (due to user programs potentially accessing deallocated pointers). The only exception being the default `heap.rs` allocator that gets enabled by a different compile time feature flag `bsan-alloc-metadata`, though this allocator uses the global allocator (and hence can also be used with different backends).

Additionally, due to the lack of garbage collection in BSAN's tree at the moment, it is difficult to run allocation heavy test programs without running out of memory (as I have become annoyingly familiar with SIGNAL 137...). For now, a simple node clearing implementation gives us insight into potential performance, though remains relatively unexplored due to the lack of OS memory return of said cleared nodes. Given our limitation in running real world crates, it is difficult to reason about our allocation layout strategies.

Running simplified test programs that stress test various data structures and access patterns (see `heavy_bench`) gives us mixed signals and remains inconclusive to allocator strategy. 

The `mimalloc` backend gives us significant performance improvements in certain programs, showing allocator performance viability. Our biggest justification for a custom allocator is the ability to control the underlying free list implementation. Specifically, it will allow us to implement frame clearing, where we give back frames to the OS (freeing physical memory) while keeping virtual pages. This will allow us to detect undefined behavior and optimize memory usage without the risk of segfaults. Additionally, there might be further performance improvements afforded by reducing redundant free lists, though this will have to be measured in a subsequent PR.

Gitter499:
# TODO
- [x] Rebase onto new `main` file structure

# Note

In order to bundle `mimalloc` as one of our backends, `bsan-script/commands.rs` was modified to pack `mimalloc` objects into our `bsan-rt` archive.

Closes #201


bsan-rt/src/global.rs
```diff
@@ -102,7 +105,10 @@ impl<'a> DerefMut for ExposedProvenanceRefMut<'a> {
 pub struct GlobalCtx {
     protected_tags: RwLock<ProtectedTags>,
     shadow_heap: ShadowHeap<Provenance>,
-    alloc_metadata_map: Heap<AllocInfo>,
+    #[cfg(feature = "alloc-bsan-metadata")]
+    alloc_metadata_heap: crate::memory::Heap<AllocInfo>,
```
icmccorm: You shouldn't need the `crate::memory` qualifier, since it's been imported above.

PR #209, opened 2026-06-01
Added support for interception.

icmccorm:
This PR adds the sanitizer common interceptors for signal handlers and syscalls. It ensures that shadow memory updates associated with `memset`, `memcpy`, and `memmove` are handled. 

This will be crucial for avoiding false positives. We cannot validate read and write accesses that occur in interceptors yet, but we can (at least optimistically) model their effects on shadow memory.

### DIFF
--- a/bsan-rt/llvm-wrapper/CMakeLists.txt
+++ b/bsan-rt/llvm-wrapper/CMakeLists.txt
@@ -33,6 +33,9 @@ append_rtti_flag(OFF BSAN_COMMON_CFLAGS)
 append_list_if(COMPILER_RT_HAS_FFREESTANDING_FLAG -ffreestanding BSAN_COMMON_CFLAGS)
 set(BSAN_DYNAMIC_CFLAGS ${BSAN_COMMON_CFLAGS})
 
+set_source_files_properties(bsan_interceptors.cpp PROPERTIES
+  COMPILE_FLAGS "-Wno-unused -Wno-unneeded-internal-declaration")
+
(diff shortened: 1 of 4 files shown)


mojeanmac (approved):
LGTM! Do we need to eventually add interceptors to other C standard memory functions like `malloc` or `memcmp`?

icmccorm:
@mojeanmac - We already have `malloc`, and `memcmp` ends up being handled through the [common interceptor macros](https://github.com/llvm/llvm-project/blob/e15e51089e6a5faef7643d8dc123788f7e91f9fc/compiler-rt/lib/sanitizer_common/sanitizer_common_interceptors.inc#L844) (e.g. `#INTERCEPTOR_READ_RANGE(..)`). 

However, these don't do everything that we would need for BorrowSanitizer. We can tell when a read access is happening, but not what provenance value should be used to validate it. This is something that we could support, but we would need to modify the common interceptor declarations. This is a "big" task, but it's mostly boilerplate. We would just need to add a parameter to each of the "READ_RANGE" and "WRITE_RANGE" calls indicating the index of the parameter whose provenance is being used (if any).

You could do this without breaking the other sanitizers by declaring something like `#INTERCEPTOR_READ_RANGE_IDX(..)`, and having this defined as calling `#INTERCEPTOR_READ_RANGE` by default, ignoring the extra parameter. Only users that override the definition of `#INTERCEPTOR_READ_RANGE_IDX(..)` would get the extra information.

Issue #192, opened 2026-05-22
Add Lazy Tree Allocation


mojeanmac:
Miri is the original Rust sanitizer to use Tree Borrows. BorrowSanitizer has copied many of its modules for borrow tracking, and recently in #182, we updated our fork of the `tree_borrows` module to be more similar to their current layout. Ours is different mainly in that we do not need to support Stacked Borrows, and we have our own interface with LLVM instead of `MiriMachine`.

Alongside Borrow Sanitizer development, I'm working on a [collaboration with Cornell](https://github.com/cmu-cornell-for-rust) towards optimizing Tree Borrows in Miri.
One finding we've uncovered is that preventing the creation of single-node trees (which consist of an average of 74% of total trees) in Miri can speedup runtime by about 5%!

I've also presented this work in the context of the Fantastic Bugs class (17712) here at CMU:
- [Slides](https://macla.ren/lazy-alloc-slides.pdf): A 5-min overview with diagrams and code snippets.

- [Report](https://macla.ren/lazy-alloc-report.pdf): The final report with some different stats, since I re-ran benchmarking with additional Miri flags and analysis.

These changes were implemented in the [`lazy-alloc`](https://github.com/cmu-cornell-for-rust/miri/tree/lazy-alloc) branch of our fork of Miri. Particularly, only [`borrow_tracker/tree_borrows/mod.rs`](https://github.com/cmu-cornell-for-rust/miri/blob/lazy-alloc/src/borrow_tracker/tree_borrows/mod.rs) and one line in [`borrow_tracker/mod.rs`](https://github.com/cmu-cornell-for-rust/miri/blob/ad2bebe7dd5e7e333bd1e56be24cb493f701f672/src/borrow_tracker/mod.rs#L257) needed to be changed. You can also see the diff in the [PR draft](https://github.com/cmu-cornell-for-rust/miri/pull/1/changes).

To make the same optimization in BorrowSanitizer, you will likely have to wrap the interface functions in [`borrow_tracker.rs`](https://github.com/BorrowSanitizer/bsan/blob/main/bsan-rt/src/borrow_tracker.rs) in our runtime and the tree functions in [`tree_borrows/tree.rs`](https://github.com/BorrowSanitizer/bsan/blob/main/bsan-rt/src/tree_borrows/tree.rs), the latter of which should be nearly identical to in Miri.

[...]

PR #210, opened 2026-06-02
Implemented Lazy Tree Allocation

vnt1c:
See #192. Implemented lazy tree allocation: defer Tree construction until first alias

### DIFF
--- a/bsan-rt/src/tree_borrows/tree.rs
+++ b/bsan-rt/src/tree_borrows/tree.rs
@@ -1186,6 +1186,141 @@ impl Node {
     }
 }
 
+/// A tree that is lazily initialized: starts as `Uninit` (storing only the root tag, size, and
+/// span needed to construct the real `Tree` on demand), and transitions to `Init` the first
+/// time a child is added via `new_child`. All operations on a single-node tree short-circuit
+/// without ever allocating the underlying `Tree`.
+#[derive(Clone, Debug)]
+pub enum LazyTree {
+    Uninit { root_tag: BorTag, size: Size, span: Span },
+    Init(Tree),
+}
+
+impl LazyTree {
+    pub fn new(root_tag: BorTag, size: Size, span: Span) -> Self {
+        LazyTree::Uninit { root_tag, size, span }
+    }
+
+    /// Forces the tree into the `Init` state, constructing the underlying `Tree` if needed.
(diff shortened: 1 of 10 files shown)


bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -1186,6 +1186,141 @@ impl Node {
...
+    Init(Tree),
+}
+
+impl LazyTree {
+    pub fn new(root_tag: BorTag, size: Size, span: Span) -> Self {
+        LazyTree::Uninit { root_tag, size, span }
+    }
+
+    /// Forces the tree into the `Init` state, constructing the underlying `Tree` if needed.
+    fn ensure_init(&mut self) -> &mut Tree {
```
mojeanmac, line 1205: There's no need to return a `Tree`, since we're already modifying `self`. That way, we don't need the `unreachable!` macros. A single `if let` check for `Uninit` should do the trick. For example in Miri I did:
```rs
pub fn ensure_initialized<'tcx>(
        &mut self,
        state: &mut GlobalStateInner,
        machine: &MiriMachine<'tcx>,
    ) {
        if let AllocState::Uninitialized { id, size, span } = self {
            let tag = state.root_ptr_tag(*id, machine);
            *self = AllocState::Initialized(Tree::new(tag, *size, *span));
        }
    }
```

bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -1186,6 +1186,141 @@ impl Node {
...
+    }
+
+    /// Marks `tag` as exposed. Forces initialization because the wildcard tracking lives
+    /// inside `Tree`. Pointer-to-integer casts are rare so the cost is acceptable.
+    pub fn expose_tag(&mut self, tag: BorTag, protected: bool) {
+        self.ensure_init().expose_tag(tag, protected);
+    }
+
+    ///// GC pass: no-op on `Uninit` trees since there are no unreachable children.
+    // pub fn remove_unreachable_tags(&mut self, live_tags: &FxHashSet<BorTag>) {
```
mojeanmac, line 1317: Feel free to leave this uncommented, just add an `#[allow(unused)]` attribute above the function signature.

mojeanmac (commented):
If you have any runtime benchmarking results, add them in the description of your next PR! Let's keep `ensure_init` and the initialization in `expose`.

PR #212, opened 2026-06-02
Add smallvec-valmap compile-time feature

### LINKED ISSUE #211: Modify `unimap.rs` to use `smallvec` internally
## Motivation

Using `smallvec` in our `Tree` -> `UniValMap` data structure allows us to save on heap allocations, subtly increasing performance and saving a respectable amount of memory. Upon differential testing, a `smallvec` with an inline buffer size of 16 achieved performance improvements across various tests. This change only showed a regression in performance in a program with many allocations (though further testing is needed).

Enabled by default, can be disabled by compiler directives

Gitter499:
Closes #211


bsan-rt/src/tree_borrows/data_structures/unimap.rs
```diff
@@ -355,4 +376,22 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn test_sizes() {
+        use crate::tree_borrows::tree::{LocationState, Node};
+        libc_print::std_name::println!("SIZEOF_NODE: {}", core::mem::size_of::<Node>());
```
icmccorm: These should be `assert!`s, not print statements.

icmccorm (changes requested):

icmccorm (approved):

PR #213, opened 2026-06-03
Wildcard Provenance (Part #1)

icmccorm:
This is step 1 for implementing wildcard provenance. 

With this PR, when we see a `ptrtoint` instruction in LLVM, we will call the function `__bsan_expose_prov` on the pointer's provenance.

This marks the borrow tag as exposed within the tree. It also adds a mapping between the bounds of the underlying allocation and its associated `AllocInfo` to a global table.
```
GlobalCtx {
  ...
  exposed_provenance: RwLock<RangeObjectMap<NonNull<AllocInfo>>>,
}
```
The `RangeObjectMap` is reused here from `StackedBorrows`. Its contents will need to be garbage-collected.

This PR also cleans up several patterns in `BorrowTracker` and `AllocInfo` that could lead to UB in our runtime. Originally, `AllocInfo` contained `Option<Mutex<Tree>>`. We would check the `Option` to validate the state of the allocation. However, liveness is allowed to be racy. This switches to using `Mutex<Option<LazyTree>>`, instead, and avoids mutably borrowing anything within `AllocInfo` that falls outside of the `Mutex`.

### DIFF
--- a/bsan-pass/BorrowSanitizer.cpp
+++ b/bsan-pass/BorrowSanitizer.cpp
@@ -1206,6 +1206,13 @@ private:
     }
   }
 
+  void visitPtrToIntInst(PtrToIntInst &I) {
+    if (auto Prov = getProvenance(I.getParent(), I.getPointerOperand())) {
+      IRBuilder<> IRB(&I);
+      IRB.CreateCall(BS.BsanFuncExposeProv, {(*Prov).Tag, (*Prov).Info});
+    }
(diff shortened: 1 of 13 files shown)


mojeanmac (approved):
I see, LGTM. So what this means for lazy allocation is that if we want to defer exposed singleton tree allocation, we'd need to add it's `AllocId` to the exposed global table.

PR #215, opened 2026-06-04
Fixed Lazy Tree Allocation

### LINKED ISSUE #192: Add Lazy Tree Allocation

vnt1c:
Fixed #192. 

Simplified ensure_init() for LazyTree, removing unreachable macros and the Tree return. Updated ensure_init() calls accordingly. Added #[allow(unused)] to LazyTree's remove_unreachable_tags


icmccorm:
We should have a newtype that we can reassign to different implementations of the `Tree`. For example,
```rust
pub type Tree = LazyTree;
```
That way, if we think of an additional variant of the `Tree`, we only have to reassign the newtype, instead of having to change all of the APIs within `BorrowTracker`.

icmccorm:
To be a bit more clear, we want to rename `Tree` to `EagerTree`, and then do the following:
```Rust
// All of the function associated with the `LazyTree` / `EagerTree` should not be 
// accessible outside of the current file. Consumers of the `AllocStateImpl` type should
// only see the methods that get defined as part of the trait `AllocState`.

// Add as features to ./bsan-rt/Cargo.toml
#[cfg(lazy)]
pub type AllocStateImpl = LazyTree;

#[cfg(eager)]
pub type AllocStateImpl = EagerTree;

pub trait AllocStateImpl { // Methods that are currently `pub` in `LazyTree`. }
impl AllocState for LazyTree {...}
impl AllocState for EagerTree {...}
```

Everywhere that `LazyTree` is being used in other interfaces, we should have `AllocStateImpl`.

bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -1193,7 +1205,7 @@ impl Node {
 #[derive(Clone, Debug)]
 pub enum LazyTree {
     Uninit { root_tag: BorTag, size: Size, span: Span },
-    Init(Tree),
+    Init(Box<EagerTree>),
```
icmccorm: Instead of a `Box`, we can just ignore the clippy warning now, with the following:
```
#[allow(clippy::large_enum_variant)]
```

bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -1339,3 +1354,182 @@ impl AccessRelatedness {
...
+        &mut self,
+        tag: BorTag,
+        access_range: AllocRange,
+        access_kind: AccessKind,
+        access_cause: AccessCause,
+        global: &GlobalState,
+        alloc_id: AllocId,
+        span: Span,
+    ) -> UBResult<()> {
+        self.perform_access(tag, access_range, access_kind, access_cause, global, alloc_id, span)
```
icmccorm: Instead, could you copy in the implementation of `self.perform_access` here (and do the same for other endpoints)?

icmccorm (changes requested):

icmccorm:
It seems like the performance got a little worse with these changes. Adding an extra level of indirection via `Box` might be hurting the runtime benefits of lazy allocation. @vnt1c could you try removing the `Box` and have `LazyTree` contain `AllocInfo` directly, and then see what that does to CI runtime?

PR #217, opened 2026-06-04
Clean our sysroot by default when running UI tests.

icmccorm:
One of the most annoying parts of our current build system is that when we run UI tests, we assume that our current sysroot is valid. Changes to the instrumentation pass or the runtime can invalidate the current sysroot build, leading to linking errors or simply weird, unpredictable behavior. 

This PR adds new default behavior for `xb ui` and `xb fix`. These commands will now remove the entire sysroot and any temporary build artifacts by default. You can pass the flag `--keep-sysroot` to either command if you want to trust these artifacts, if they exist.


mojeanmac (approved):
This is nice, so we no longer should need to delete `target` when modifying cpp :)

PR #220, opened 2026-06-06
Merge all dependencies of the `BorrowSanitizer` pass into a single file

icmccorm:
This PR merges `Provenance.cpp` and parts of `BorrowSanitizerPass.cpp` into `BorrowSanitizer.cpp`. This monolithic structure matches other LLVM sanitizer passes. It is also helpful for supporting a fixed shadow memory mapping. The shadow memory layout will be "owned" by the module-level `BorrowSanitizer` struct, which `Provenance` objects will need to access to know where to load and store from.

This also improves documentation a bit.


icmccorm:
Note: we still want retags to be defined separately, since we need them for our separate static analysis pass.

mojeanmac (approved):
LGTM

PR #221, opened 2026-06-06
Switch to using the sanitizers' allocator implementation for system allocations.

icmccorm:
This is the first step for implementing #72. It adds fixed memory layouts for ARM and x86. These only specify where application memory lives. This includes the stack, global variables, the binary itself, and a region for heap allocations. You can view the layout in [`bsan_shadow.h`](https://github.com/BorrowSanitizer/bsan/blob/bdfae928fd29ffb04b18e20d28ebe3f350cb0a5d/bsan-rt/llvm-wrapper/bsan_shadow.h).

We ensure that heap allocations are placed in a fixed region by switching to using the sanitizers' system allocator. The allocator is essentially boilerplate—it matches what MemorySanitizer and DataFlowSanitizer use.

We also need to do some additional work to avoid reentry into interceptors from the Rust side of things. Any API endpoint should be called *after* we declare a barrier to interception.
```C++
InterceptorBarrier barrier;
```
The only exception to this is endpoints that are only ever called after this barrier has been placed. That includes `__bsan_alloc` and `__bsan_dealloc`.

Finally, the test [`address_reuse.rs`](https://github.com/BorrowSanitizer/bsan/blob/bdfae928fd29ffb04b18e20d28ebe3f350cb0a5d/tests/miri-tests/excluded/pass/address-reuse.rs) has been moved into the `excluded` folder. This test asserted that repeated heap allocations and deallocations of the same size would eventually use a different address. This happened with malloc, but not for the sanitizer allocator, and the test is really only relevant for Miri. It might be helpful for us to avoid reusing addresses, but it isn't an essential feature of our tool.


icmccorm:
@Gitter499 I added a fixed region for your custom allocator (see #223).
```
const uptr kMetaAllocatorSpace = 0x100000000000ULL;
const uptr kMetaAllocatorSpaceSize = 0x10000000000ULL; // 1T.
```
We can implement this in `bsan_meta_allocator.h/cpp`.

PR #222, opened 2026-06-07
Add `--allow-unsafe-deps` flag for testing.

icmccorm:
This adds a new flag `--allow-unsafe-deps` to `xb ui` and `xb fix`. Tests with dependencies will not be built unless this flag is set. 

This is necessary because we want to include crates with undefined behavior or vulnerable dependencies in our UI tests, so that we can recreate the errors with BorrowSanitizer. Since these crates are only used for testing purposes, there isn't as much of a risk, but we still do not want to build them in our default configuration. Contributors should explicitly consent to this risk, even though it's pretty negligible.

These tests will run in CI. The PR also modifies our GitHub actions script for CI to further limit the permissions associated with the GitHub token, just in case.


mojeanmac (approved):
LGTM!

PR #227, opened 2026-06-09
Add random shuffling to bytewise copy test

icmccorm:
Randomly selects the order in which bytes are copied for our byte-level provenance failing test.


Gitter499 (approved):
Compiler can't optimize anything if the shuffling is dependent on the runtime randomness! Clever

PR #231, opened 2026-06-12
Relevant Line Reporting in Diagnostics

### LINKED ISSUE #186: Ensure helpful stack traces and diagnostic messages
This PR has three main issues to tackle regarding error messages:

## 1. Fix stacktrace length flag

- [x] ~~`BSAN_FLAG=stacktrace_max_len=...` as added in #161 seems to be broken. Setting it manually on line 66 of `bsan.h` still works, so it must be something on the flag processing side.~~

Correction: it was always `BSAN_OPTIONS=stacktrace_max_len=...`

## 2. Smart stacktrace unrolling

In `fail/aliasing/bsan-ptr-craft.rs`, we get the following trace by default:
```
stack backtrace:
0: <std::alloc::System as core::alloc::global::GlobalAlloc>::dealloc
      at /root/.rustup/toolchains/bsan/lib/rustlib/src/rust/library/std/src/sys/alloc/unix.rs:48:18
1: __rustc::__rdl_dealloc
      at /root/.rustup/toolchains/bsan/lib/rustlib/src/rust/library/std/src/alloc.rs:464:6
2: alloc::alloc::dealloc_nonnull
      at /root/.rustup/toolchains/bsan/lib/rustlib/src/rust/library/alloc/src/alloc.rs:128:14
```
However, if we unrolled it a little more, we can actually see where in our program the diagnostic originates:
```
...
9: bsan_ptr_craft::foo
      at /workspaces/bsan/tests/fail/aliasing/bsan-ptr-craft.rs:9:9
10: bsan_ptr_craft::main
      at /workspaces/bsan/tests/fail/aliasing/bsan-ptr-craft.rs:14:5
11: <fn() as core::ops::function::FnOnce<()>>::call_once
      at /root/.rustup/toolchains/bsan/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
12: std::sys::backtrace::__rust_begin_short_backtrace::<fn(), ()>
      at /root/.rustup/toolchains/bsan/lib/rustlib/src/rust/library/std/src/sys/backtrace.rs:166:18
```
- [x] What if by default, we automatically unrolled it until we hit the user file or `std/src/sys/backtrace.rs`?

## 3. Relevant source location resolution

Furthermore, the first error message for `bsan-ptr-craft.rs` is unhelpful as it points to a standard library location:
```
error: Undefined Behavior: deallocation through <1117>(WeakProtector) at alloc282[0x0] is forbidden
    --> /root/.rustup/toolchains/bsan/lib/rustlib/src/rust/library/std/src/sys/alloc/unix.rs:48:18
     |
  48 | unsafe { libc::free(ptr as *mut libc::c_void) }
     |
```
This line matches index `0` in the stack backtrace, but we would rather have it display the source of index `9` as shown above, which resolves to:
```rs
drop(b2);
```
- [ ] In cases where the source location is in a standard library file, can we follow the stacktrace down to the first relevant source location and display that instead?

mojeanmac:
This PR improves diagnostic reporting by retrieving the user's source code, identified as the first source path that is not library code (in cargo or rustup) as the first displayed message.

If user code is found, the library source is still included as a note immediately below the new source location, e.g:
```
error: Undefined Behavior: write access through <880>(StrongProtector) at alloc155[0x0] is forbidden
   --> /workspaces/bsan/tests/miri-tests/fail/aliasing/both_borrows/aliasing_mut4.rs:11:7
    |
 11 | y.set(1);
    |
    = note: the above line calls library code, where the error was detected:
   --> /root/.rustup/toolchains/bsan/lib/rustlib/src/rust/library/core/src/mem/mod.rs:931:49
    |
931 | crate::intrinsics::write_via_move(dest, src);
    |
```

This ensures that our error messages are more useful and closer to (and in some cases more helpful than) Miri's.

~~**Performance impact**: node size increased by `+8n` bytes to account for `n` additional 8 byte span pcs~~ for this, we should instead have a flag for if the user wants to resolve historical locations, since this can significantly increase node size.

Minor formatting fixes were also added:
- `help: this transition...` messages are now indented with `=help` as they correspond to the message above them
- added a note that the stack trace is automatically truncated to 3 frames, with the `BSAN_OPTIONS` flag to extend it if warranted


icmccorm:
As we discussed earlier, we should either print both the exact source location and the user-relevant one, or add a runtime flag to enable precise printing, just in case there ends up being a situation where we need the full information.

bsan-rt/llvm-wrapper/bsan.cpp
```diff
@@ -99,6 +111,47 @@ void PrintStackTrace(StackTrace &stack) {
       frame_desc.clear();
     }
   }
+  MaybeWarnTruncated(stack);
+}
+
+// Returns true if the file path belongs to a cargo or rustup library.
+static bool IsLibraryFile(const char *file) {
```
icmccorm, line 167: Can we have these strings in a static list, and then iterate over it? I'm thinking that C/C++ dependencies might introduce other library paths that we might want special handling for.

PR #236, opened 2026-06-15
Implemented `RefCount` struct with appropriate endpoints.

### LINKED ISSUE #233: Implement reference counting structure and related endpoints
Prerequisite to #196. The LLVM pass now emits calls to `__bsan_rc_inc` and `__bsan_rc_dec`. The Rust stubs currently do nothing. Thus, this issue covers filling them with actual reference counting logic:

1. Add a RefCount struct with correct atomic semantics (`Relaxed` increment, `Release` decrement, `Acquire` fence on reaching zero). 
2. Implement foundational functions for `RefCount`: `increment()`, `decrement()`, `get()`, and `is_unique()`.

vnt1c:
Tests delivered in [refcount.rs]. Closes #233.

* Single-threaded: new starts at 1, increment/decrement/get work as expected, decrement returns true only when hitting zero.
* Concurrent: 16 threads do balanced inc/dec ops and the count returns to 1. Exactly one thread observes zero on the final decrement. A Release/Acquire test verifies that writes from all threads are visible to the thread that hits zero (Miri catches ordering violations here). Thread counts scale down under cfg!(miri).

Results.

* `cargo test`: 9/9 pass.
* `cargo miri test`: 9/9 pass, and the three concurrent tests pass across 16 Miri seeds. No ordering violations found.

### DIFF
--- a/bsan-rt/src/tree_borrows/refcount.rs
+++ b/bsan-rt/src/tree_borrows/refcount.rs
@@ -5,15 +5,14 @@ use core::sync::atomic::{fence, AtomicUsize, Ordering};
 // endpoints (lib.rs) actually drive this type.
 #[allow(dead_code)]
 #[derive(Debug)]
-pub struct RefCount {
-    count: AtomicUsize,
-}
+#[repr(transparent)]
+pub struct RefCount(AtomicUsize);
 
 #[allow(dead_code)]
 impl RefCount {
     /// Creates a new `RefCount` initialized to 1.
     pub fn new() -> Self {
-        Self { count: AtomicUsize::new(1) }
+        Self(AtomicUsize::new(1))
     }
 
     /// Increments the reference count.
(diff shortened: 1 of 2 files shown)


bsan-rt/src/tree_borrows/refcount.rs
```diff
@@ -0,0 +1,234 @@
...
+#[cfg(test)]
+mod tests {
+    use core::cell::UnsafeCell;
+    use std::sync::atomic::{AtomicUsize, Ordering};
+    use std::sync::Arc;
+    use std::thread;
+
+    use super::*;
+
+    // ---- single-threaded API sanity checks ----
```
icmccorm:
```suggestion
    // ---- single-threaded API smoke checks ----
```

bsan-rt/src/tree_borrows/refcount.rs
```diff
@@ -0,0 +1,234 @@
+use core::sync::atomic::{fence, AtomicUsize, Ordering};
+
+/// A thread-safe reference count with correct atomic ordering semantics.
+// TODO: remove `allow(dead_code)` once the `__bsan_rc_inc`/`__bsan_rc_dec`
+// endpoints (lib.rs) actually drive this type.
+#[allow(dead_code)]
+#[derive(Debug)]
+pub struct RefCount {
```
icmccorm: For structs like this, I'm more a fan of the following pattern:
```Rust
#[derive(Debug)]
#[repr(transparent)]
pub struct RefCount(AtomicUsize);

impl RefCount {
    /// Creates a new `RefCount` initialized to 1.
    pub fn new() -> Self {
        Self(AtomicUsize::new(1))
    }
}
```

icmccorm (approved):
LGTM

PR #239, opened 2026-06-16
Implemented relevant functions for reference counting for nodes

### LINKED ISSUE #237: Implement reference counting APIs
Here's what remains for implementing reference counting. 

1. Add a reference count to every [`Node`](https://github.com/BorrowSanitizer/bsan/blob/d217edabaaa78929403d961d7ae9a6afcdb4f29f/bsan-rt/src/tree_borrows/tree.rs#L317)
2. Add a new API to [`AllocState`](https://github.com/BorrowSanitizer/bsan/blob/d217edabaaa78929403d961d7ae9a6afcdb4f29f/bsan-rt/src/tree_borrows/tree.rs#L944) that returns a reference to a `Node`.
3. Add increment and decrement APIs to [`BorrowTracker`](https://github.com/BorrowSanitizer/bsan/blob/d217edabaaa78929403d961d7ae9a6afcdb4f29f/bsan-rt/src/borrow_tracker.rs#L18). These should use the API implemented in the previous step to update the reference count for the `Node` that is associated with the `BorTag` field of `BorrowTracker`
4. Within `__bsan_rc_inc` and `__bsan_rc_dec`, use `BorrowTracker::for_alloc` (take a look at [`__bsan_protector_end`](https://github.com/BorrowSanitizer/bsan/blob/d217edabaaa78929403d961d7ae9a6afcdb4f29f/bsan-rt/src/lib.rs#L469) for an example) to create an instance of `BorrowTracker` and call the corresponding increment and decrement APIs.

We'll handle the case when a reference count hits zero from within the `BorrowTracker`.

vnt1c:
Closes #237. Adds [`RefCount`](https://github.com/BorrowSanitizer/bsan/blob/2ebcb3c7dc5214e2e4840b41cb92a689ee6f4c87/bsan-rt/src/tree_borrows/refcount.rs#L11) fields to every `Node` (stored in `EagerTree`) and `LazyTree`'s `Uninit` variant (the lazy root). Implements `increment`/`decrement` on the `AllocState` trait for both `EagerTree` and `LazyTree` (the lazy impl updates the `Uninit` count without forcing tree initialization) and exposes them on `BorrowTracker`, which the `__bsan_rc_inc_impl`/`__bsan_rc_dec_impl` endpoints call.

Note that the zero-transition is currently unhandled; `decrement()` returns true when the count hits zero, but the endpoint discards it for now. Cleanup when reference counts hit zero should be implemented later from within `BorrowTracker`. Additionally, the instrumentation pass does not call these functions yet either.


bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -1078,6 +1088,22 @@ impl AllocState for LazyTree {
             tree.remove_unreachable_tags(live_tags);
         }
     }
+    fn increment(&self, tag: BorTag) {
+        match self {
+            LazyTree::Uninit { root_tag, refcount, .. } => {
+                if *root_tag == tag {
```
icmccorm: If our tree is `LazyTree::Uninit`, then `root_tag` will always be equal to `tag`. We should add this as a debug assertion.

bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -1078,6 +1088,22 @@ impl AllocState for LazyTree {
...
+                    refcount.increment();
+                }
+            }
+            LazyTree::Init(tree) => tree.increment(tag),
+        }
+    }
+    fn decrement(&self, tag: BorTag) -> bool {
+        match self {
+            LazyTree::Uninit { root_tag, refcount, .. } => *root_tag == tag && refcount.decrement(),
```
icmccorm: Likewise, we don't need to reassign `root_tag` here. If we reach this step, then we know for sure that the tag is some concrete value outside of the range of reserved tags for wildcard / omnivalid / null provenance.

bsan-rt/src/lib.rs
```diff
@@ -589,11 +589,20 @@ unsafe extern "C-unwind" fn __bsan_shadow(addr: *mut c_void) -> NonNull<Provenan
 
 /// Increments the reference count associated with a provenance value.
 #[unsafe(no_mangle)]
-unsafe extern "C-unwind" fn __bsan_rc_inc_impl(_bor_tag: BorTag, _alloc_info: *mut AllocInfo) {}
+unsafe extern "C-unwind" fn __bsan_rc_inc_impl(bor_tag: BorTag, alloc_info: *mut AllocInfo) {
+    let prov = Provenance { bor_tag, alloc_info };
+    let _ = BorrowTracker::for_alloc(prov, |bt| {
+        bt.increment();
```
icmccorm: If you have increment and decrement return a `UBResult<()>`, then you won't need the extra `Ok` here and in the body of `rc_dec_impl`

bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -334,6 +335,9 @@ pub struct Node {
     default_initial_idempotent_foreign_access: IdempotentForeignAccess,
     /// Whether a wildcard access could happen through this node.
     pub is_exposed: bool,
+    /// Number of live references to this node. Always accessed under the
```
icmccorm, line 338: For this reason, we will not need atomic reference count updates for the moment. In the future though, we'll want these to be atomic, so that we can allow multiple threads to access individual nodes.

For now, could you add the following APIs:
```Rust
    pub fn increment_nonatomic(&self) {
        let prev = unsafe {
            *self.0.as_ptr() += 1;
        };
    }
    
    pub fn decrement_nonatomic(&self) {
        let prev = unsafe {
            *self.0.as_ptr() -= 1;
        };
    }
```
Or something similar? We can use these instead.

icmccorm (changes requested):

PR #243, opened 2026-06-17
Update benchmarking script to add an "all checks enabled" mode and compare against Miri

icmccorm:
This updates our benchmarking script to include our full runtime checking mode and to compare against Miri. 

It also bumps the version of our custom Rust fork. The newest changes make it so that we skip parsing terminal information in `libtest` and use default values instead, if we have `--cfg=bsan` enabled. Miri also does this under `--cfg=miri`, because this parsing step is expensive to validate.

### DIFF
--- a/bsan-script/etc/bench.py
+++ b/bsan-script/etc/bench.py
@@ -22,14 +22,16 @@ NATIVE = {
 
 MIRI = {
     "name": "miri-tb",
-    "flags": [
-        "-Zmiri-tree-borrows",
-        "-Zmiri-provenance-gc=0",
-        "-Zmiri-mute-stdout-stderr",
-        "-Zmiri-disable-data-race-detector",
-        "-Zmiri-deterministic-concurrency",
-        "-Zmiri-disable-alignment-check"
-    ],
+    "env": {
+        "MIRIFLAGS": [
+            "-Zmiri-tree-borrows",
+            "-Zmiri-provenance-gc=0",
+            "-Zmiri-mute-stdout-stderr",
+            "-Zmiri-disable-data-race-detector",
+            "-Zmiri-deterministic-concurrency",
+            "-Zmiri-disable-alignment-check"
+        ]
+    },
     "runs": 3,
     "warmup": 1
 }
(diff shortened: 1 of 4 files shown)


bsan-script/etc/bench.py
```diff
@@ -104,11 +168,85 @@ def compile_test_binary(cargo_cmd: list[str], label: str, cwd: Path, out: Path)
...
+    final `miri` invocation as a standalone script."""
+    replay = scratch / "miri-replay.sh"
+
+    override_env = { **(config.get("env") or {}) }
+    override_env["MIRI"] = str(ensure_miri_wrapper(scratch))
+    override_env["BSAN_MIRI_REPLAY"] = str(replay)
+
+    # One cargo-miri run generates the replay script for this test's final miri
+    # invocation; we then time that script on its own, free of cargo overhead.
+    run_capture(["cargo", "miri", "test", "-q", "--lib", "--", "--exact", t, "--nocapture"],
```
mojeanmac, line 245: It's not clear whether we're passing the same arguments to the bsan execution. The dedicated miri execution functions seem a bit redundant-- isn't the only difference the Zmiri flags and `cargo miri`? Also, passing `--lib` typically skips a lot of tests, so I'm not sure if it was intentional or not (it may not make a difference for these two crates).

icmccorm: > It's not clear whether we're passing the same arguments to the bsan execution.

We pass `--exact {test name} --nocapture` to both Miri and each of the bsan / native execution variants. The `--lib` flag only goes to Cargo.

> The dedicated miri execution functions seem a bit redundant-- isn't the only difference the Zmiri flags and cargo miri?

I suppose we could refactor this so that the shell script we emit to replicate the final invocation of Miri would be executed as if it were one of the native binaries. 

> Also, passing --lib typically skips a lot of tests, so I'm not sure if it was intentional or not (it may not make a difference for these two crates).

I typically skip doctests under the entirely vibes-based assumption that any crate with doctests will also have unit tests with the same coverage. I have no idea if that's anywhere near true, though. Honestly I think someone needs to do an empirical study along the lines of "when do Rust programmers use doctests?"

mojeanmac: Ok, LGTM

Issue #238, opened 2026-06-16
Increment and decrement reference counts as a side effect of memintrinsics


icmccorm:
In addition to the regular `load` and `store` operations that we handle using our instrumentation pass, we need to increment and decrement reference counts when we call memory intrinsics. This includes `memset`, `memcpy`, `memmove`, etc. We abstract over these intrinsics using [several helper functions](https://github.com/BorrowSanitizer/bsan/blob/157756994c54369bf04ebe7365b9c626a573145f/bsan-rt/llvm-wrapper/bsan_shadow.cpp#L180) inside the LLVM runtime.

When we clear provenance using `ClearShadow`, we want to decrement the reference count of every provenance value. When we `MoveShadow` or `CopyShadow`, we want to increment the replacement, and decrement the provenance being replaced. But, it's crucial that we avoid performing redundant reference count operations when the source and destination of a `MoveShadow` are overlapping. Also, we only want to load the borrow tag of a provenance value to modify its reference count if we are certain that its `AllocInfo *` component represents a valid allocation (e.g. it's not a null pointer).

PR #244, opened 2026-06-17
Decrement reference count for cleared shadow values

Gitter499:
Closes first part of #238

### DIFF
--- a/bsan-rt/llvm-wrapper/bsan_shadow.cpp
+++ b/bsan-rt/llvm-wrapper/bsan_shadow.cpp
@@ -235,6 +235,7 @@ void ClearShadow(void *dest, uptr size) {
       __bsan_rc_dec(*tag_ptr, *info_ptr);
 
     *tag_ptr = 0;
+    *info_ptr = nullptr;
   }
 }
 
--- a/bsan-rt/llvm-wrapper/bsan_shadow.cpp
+++ b/bsan-rt/llvm-wrapper/bsan_shadow.cpp
@@ -234,7 +234,6 @@ void ClearShadow(void *dest, uptr size) {
     if (*info_ptr != nullptr)
       __bsan_rc_dec(*tag_ptr, *info_ptr);
 
-    *tag_ptr = 0;
     *info_ptr = nullptr;
   }
 }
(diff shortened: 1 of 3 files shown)


bsan-rt/llvm-wrapper/bsan_shadow.cpp
```diff
@@ -220,7 +220,20 @@ void ClearShadow(void *dest, uptr size) {
     return;
   uptr d_aligned, d_size;
   AlignRange8((uptr)dest, size, d_aligned, d_size);
-  internal_memset((void *)MEM_TO_SHADOW(d_aligned), 0, d_size);
-  internal_memset((void *)MEM_TO_ORIGIN(d_aligned), 0, d_size);
+
+  uptr shadow_start = MEM_TO_SHADOW(d_aligned);
+  uptr origin_start = MEM_TO_ORIGIN(d_aligned);
+
+  const uptr step = 8; // kMinProvAlignment
```
icmccorm: This should just be `kMinProvAlignment`.

bsan-rt/llvm-wrapper/bsan_shadow.cpp
```diff
@@ -220,7 +220,20 @@ void ClearShadow(void *dest, uptr size) {
...
+
+  for (uptr s = d_aligned; s < d_aligned + d_size; s += step) {
+    uptr offset = s - d_aligned;
```
icmccorm: * We only want to load `AllocInfo *` and call `__bsan_rc_dec` if the `BorTag` is nonzero, or vice-versa.

* We want to merge and inline the calls to `internal_memset` into the body of the `for` loop. As we iterate over each provenance value, we should also zero-out the corresponding section of memory after calling `__bsan_rc_dec`. That way, we avoid iterating over the relevant sections of memory three times (once for decrements, and twice for the calls to `internal_memset`)

icmccorm (changes requested):

bsan-rt/llvm-wrapper/bsan_shadow.cpp
```diff
@@ -220,7 +220,22 @@ void ClearShadow(void *dest, uptr size) {
...
+
+  for (uptr offset = 0; offset < d_size; offset += step) {
+    BorTag *tag_ptr = reinterpret_cast<BorTag *>(shadow_start + offset);
+    AllocInfo **info_ptr =
+        reinterpret_cast<AllocInfo **>(origin_start + offset);
+
+    if (*info_ptr != nullptr)
+      __bsan_rc_dec(*tag_ptr, *info_ptr);
+
+    *tag_ptr = 0;
```
Gitter499: Do we need to set this to 0?

icmccorm: Actually, no! We can skip clearing tags.

Issue #241, opened 2026-06-17
Implement support for pruning "dead" tags from a `Tree`


icmccorm:
Miri's Tree Borrows implementation has mechanisms for garbage collecting a particular `Tree`. These are all invoked via the function `remove_unreachable_tags`:

```Rust
fn remove_unreachable_tags(&mut self, live_tags: &FxHashSet<BorTag>) {
    for i in 0..(self.roots.len()) {
        self.remove_useless_children(self.roots[i], live_tags);
    }
    self.locations.merge_adjacent_thorough();
}
```

This is designed around the assumption that we will have a tracing collector, where we collect the global set of "live" borrow tags that are still reachable in memory and remove any state associated with tags in the `Tree` that are not stored within this set, indicating that they are no longer reachable.

However, our garbage collector will be oriented toward reference counting. We may add some form of tracing (e.g. "deferred" reference counting), but our perspective will still be the inverse of Miri's. Instead of a set of "live" tags, we will have a set of "dead" tags that need to be removed.

We need to adjust Miri's implementation to support removing a set of "dead" tags from the tree. This implementation should coexist with the current "live" tags implementation—we don't want to remove what's already there, we just want to add an additional collection mechanism that takes a set of "dead" tags instead of a set of "live" ones.

icmccorm:
Finished in #245

PR #245, opened 2026-06-18
Dead list versions of tree GC functions

vnt1c:
Relevant to #241. Created dead list versions of the existing Miri tree GC functions.


bsan-pass/BorrowSanitizer.cpp
```diff
@@ -586,10 +586,10 @@ void BorrowSanitizer::initializeCallbacks(Module &M,
   BsanFuncValidateRetval = M.getOrInsertFunction(
       BSAN("validate_retval"), AL, IRB.getVoidTy(), PtrTy, PtrTy, IntptrTy);
```
icmccorm, line 588: We don't need to have this return `i1`. Only the innermost implementation, the "`impl`" variants, will return `bool`.

bsan-pass/BorrowSanitizer.cpp
```diff
@@ -586,10 +586,10 @@ void BorrowSanitizer::initializeCallbacks(Module &M,
...
 
-  BsanFuncRcInc = M.getOrInsertFunction(BSAN("rc_inc"), AL, IRB.getVoidTy(),
+  BsanFuncRcInc = M.getOrInsertFunction(BSAN("rc_inc"), AL, IRB.getInt1Ty(),
```
icmccorm: This should still be `IRB.getVoidTy()`.

bsan-rt/llvm-wrapper/bsan.cpp
```diff
@@ -345,24 +345,28 @@ void __bsan_write(void *ptr, uptr access_size, BorTag bor_tag,
   }
 }
 
-SANITIZER_WEAK_ATTRIBUTE void __bsan_rc_inc_impl(BorTag Tag, AllocInfo *Info);
+SANITIZER_WEAK_ATTRIBUTE
+bool __bsan_rc_inc_impl(BorTag Tag, AllocInfo *Info);
+
 SANITIZER_INTERFACE_ATTRIBUTE
-void __bsan_rc_inc(BorTag Tag, AllocInfo *Info) {
+bool __bsan_rc_inc(BorTag Tag, AllocInfo *Info) {
```
icmccorm: This should also be `void`. However, each of the `_impl` variants should still return a `bool` from the Rust side.

bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -578,6 +627,68 @@ impl EagerTree {
             }
         }
     }
+
+    /// Like [`Self::remove_useless_children`], but takes a set of *dead* tags
+    /// to remove instead of a set of *live* tags to keep. See [`AllocState::remove_dead_tags`].
+    fn remove_useless_children_dead(&mut self, root: UniIndex, dead: &FxHashSet<BorTag>) {
```
icmccorm, line 633: This is correct, but we can be much more efficient. We should iterate over the set of dead nodes and remove each one. You'll still need to merge children into a parent, if a dead node still has reachable children. But, we shouldn't need the full DFS.

icmccorm (changes requested):

PR #247, opened 2026-06-18
Implement `CopyShadow` and `MoveShadow` RC

### LINKED ISSUE #238: Increment and decrement reference counts as a side effect of memintrinsics

Gitter499:
Closes #238

### DIFF
--- a/bsan-rt/llvm-wrapper/bsan_shadow.cpp
+++ b/bsan-rt/llvm-wrapper/bsan_shadow.cpp
@@ -194,7 +194,8 @@ void CopyAligned(void *dest, const void *src, uptr size) {
   internal_memcpy((void *)d_aligned, (const void *)s_aligned, s_size);
 }
 
-static void UpdateShadowSlot(uptr d_aligned, uptr s_aligned, uptr offset) {
+ALWAYS_INLINE static void UpdateShadowSlot(uptr d_aligned, uptr s_aligned,
+                                           uptr offset) {
   uptr d_shadow = MEM_TO_SHADOW(d_aligned);
   uptr d_origin = MEM_TO_ORIGIN(d_aligned);
   uptr s_shadow = MEM_TO_SHADOW(s_aligned);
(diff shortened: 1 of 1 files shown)


bsan-rt/llvm-wrapper/bsan_shadow.cpp
```diff
@@ -190,29 +190,82 @@ void CopyAligned(void *dest, const void *src, uptr size) {
...
+  BorTag *dest_tag = reinterpret_cast<BorTag *>(d_shadow + offset);
+  BorTag *source_tag = reinterpret_cast<BorTag *>(s_shadow + offset);
+
+  if (*dest_info != nullptr)
+    __bsan_rc_dec(*dest_tag, *dest_info);
+
+  if (*source_info != nullptr)
+    __bsan_rc_inc(*source_tag, *source_info);
+
+  for (uptr i = 0; i < kMinProvAlignment; ++i) {
```
icmccorm: This can be done as a byte-level copy without the for-loop. We can think of provenance as being two parallel, 8-byte chunks in our "shadow" and "tag" regions.

bsan-rt/llvm-wrapper/bsan_shadow.cpp
```diff
@@ -190,29 +190,82 @@ void CopyAligned(void *dest, const void *src, uptr size) {
...
-  CopyAligned((void *)MEM_TO_ORIGIN(dest), (const void *)MEM_TO_ORIGIN(src),
-              size);
+
+  const uptr step = kMinProvAlignment;
+
+  uptr d_aligned;
+  uptr s_aligned, s_size;
+  AlignPtr8((uptr)dest, d_aligned);
+  AlignRange8((uptr)src, size, s_aligned, s_size);
+  if (s_size == 0)
```
icmccorm: Let's move this up to match the other early return statements.

bsan-rt/llvm-wrapper/bsan_shadow.cpp
```diff
@@ -190,29 +190,82 @@ void CopyAligned(void *dest, const void *src, uptr size) {
...
+    return;
+
+  for (uptr offset = 0; offset < s_size; offset += step)
+    UpdateShadowSlot(d_aligned, s_aligned, offset);
```
icmccorm, line 239: We should add an inlining annotation on `UpdateShadowSlot`. I believe that LLVM's compiler-rt supports `INLINE`, but I'm unsure.

bsan-rt/llvm-wrapper/bsan_shadow.cpp
icmccorm, line 239:
```suggestion
if (*info_ptr != nullptr) {
  __bsan_rc_dec(*tag_ptr, *info_ptr);
  *info_ptr = nullptr;
}
```

icmccorm: We never need to reset the tag, because all provenance validation depends on the `AllocInfo*` value. We only need to reset the `AllocInfo*` if we know that it's non-null.

icmccorm (changes requested):

icmccorm (approved):

Issue #255, opened 2026-06-25
Better Tree Pruning


vnt1c:
Currently, there is room for optimization within the tree implementations. A lot of these functions were ported over from Miri, and still use logic based off of Miri's tracing system. Miri does a DFS from the root. There are potential possibilities of doing a DFS from the nodes in the `dead` list OR doing a reserve traversal through the `dead` list under the assumption that it is already sorted.

Additionally, Miri only handles two cases: leaf nodes and nodes that have exactly one child. Otherwise, it does not free the node. This is also something to consider.

icmccorm:
Note that we'll still need to have some kind of depth first-traversal. However, instead of starting at the root, we can start at the nodes in the set. 

Another thing to note: over in the GC branch, we provide the pruning function with a list of tags, which is then converted to a `FxHashSet`. This list is sorted and guaranteed not to contain any duplicates. You may want to iterate over tags from the least to the greatest, since larger tags are children or foreign to smaller tags.

icmccorm:
Finished with #288

PR #256, opened 2026-06-25
Re-implemented tree pruning for GC

vnt1c:
Relevant to #255. Re-implemented `remove_useless_children_dead` with one pass rather than a DFS. Note that we are temporarily converting the `dead_tags` hashset into an array slice to pass the assertion. Once the GC is merged to main, we can remove this sort. While more efficient, this function requires that the dead_tags array is in ascending order (by tag). It is recommended that we heavily test this to ensure this invariant is upheld when testing the GC.

The old algorithm is `O(N)` since we visit every node in the tree once while this algorithm is `O(D)` where D is the number of dead tags. Not sure how much this will optimize GC but it should not result in a slower execution time for the GC pass.

### DIFF
--- a/bsan-rt/src/tree_borrows/tree.rs
+++ b/bsan-rt/src/tree_borrows/tree.rs
@@ -679,6 +679,9 @@ impl EagerTree {
                 self.remove_useless_node(idx);
             }
             // Otherwise, the dead node has more than one reachable child and we retain it
+
+            // To-do: Consider cases where there are multiple descendants of a parent
+            // node with varying permissions
         }
     }
 }
--- a/bsan-rt/src/tree_borrows/tree.rs
+++ b/bsan-rt/src/tree_borrows/tree.rs
@@ -497,8 +497,6 @@ impl EagerTree {
 
         let [child_idx] = node.children[..] else { return None };
 
-        // We never want to replace the root node, as it is also kept in `root_ptr_tags`.
-        node.parent?;
         // Since protected nodes are never GC'd (see `borrow_tracker::FrameExtra::visit_provenance`),
         // they are never added to the dead set, so reaching this point (where `node.tag` is
         // in `dead`) guarantees that `node` is not protected.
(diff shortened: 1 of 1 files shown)


bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -448,13 +448,6 @@ impl EagerTree {
         node.children.is_empty() && !live.contains(&node.tag)
     }
 
-    /// Like [`Self::is_useless`], but takes a set of *dead* tags instead of
-    /// *live* ones.
-    fn is_useless_dead(&self, idx: UniIndex, dead: &FxHashSet<BorTag>) -> bool {
```
vnt1c, line 453: I realized we do not need this since we do not need to check if the dead list contains the tag. We simply check `node.children.is_empty()` in the `remove_useless_children_dead()` function

bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -628,65 +614,46 @@ impl EagerTree {
...
+            dead_tags.is_sorted(),
+            "remove_useless_children_dead requires ascending-sorted tags"
+        );
 
-                // We are done, the parent can continue.
-                stack.pop();
-                continue;
+        // Iterating through dead_tags in reverse (descending tag order)
+        for &tag in dead_tags.iter().rev() {
+            // Skip tags belonging to other allocations / already-removed nodes.
```
icmccorm: This comment should be reworded. All of the tags are guaranteed to have belonged to this allocation. It is possible that we will see a tag that has already been removed.

bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -628,65 +614,46 @@ impl EagerTree {
...
+            let Some(idx) = self.tag_mapping.get(&tag) else { continue };
+            let node = self.nodes.get(idx).unwrap();
+            // Never remove a root (it is also tracked in `root_ptr_tags`).
```
icmccorm: We can remove the root.

bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -496,23 +489,16 @@ impl EagerTree {
...
+    fn can_be_replaced_by_single_child_dead(&self, idx: UniIndex) -> Option<UniIndex> {
         let node = self.nodes.get(idx).unwrap();
 
         let [child_idx] = node.children[..] else { return None };
 
         // We never want to replace the root node, as it is also kept in `root_ptr_tags`.
-        if !dead.contains(&node.tag) || node.parent.is_none() {
-            return None;
-        }
+        node.parent?;
```
icmccorm: We can remove the root, which will trigger deallocating the entire allocation metadata. We should have special handling for this. If we remove a root node, then remove it from `root_ptr_tags`, too.

bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -628,65 +614,46 @@ impl EagerTree {
...
+        debug_assert!(
+            dead_tags.is_sorted(),
+            "remove_useless_children_dead requires ascending-sorted tags"
+        );
 
-                // We are done, the parent can continue.
-                stack.pop();
-                continue;
+        // Iterating through dead_tags in reverse (descending tag order)
+        for &tag in dead_tags.iter().rev() {
```
icmccorm, line 631: It is possible that a node will have a nonzero reference count at this stage. Our ZCT is imprecise. We need to do a final check for that here, before we can remove a node.

bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -1495,9 +1462,14 @@ impl AllocState for EagerTree {
         self.locations.merge_adjacent_thorough();
     }
     fn remove_dead_tags(&mut self, dead_tags: &FxHashSet<BorTag>) {
-        for i in 0..(self.roots.len()) {
-            self.remove_useless_children_dead(self.roots[i], dead_tags);
-        }
+        // `remove_useless_children_dead` consumes the dead tags in ascending order
```
icmccorm: At this point, let's just have `remove_dead_tags` receive `&[BorTag]`.

icmccorm (changes requested):

icmccorm:
@vnt1c there's one additional change that we need here. Every node should have a "dead" flag, which must set in situations where we identify a dead node with more than one live child. Future GC runs should treat a node as dead if it exists in the list of dead tags, or if it has its flag set. 

Our garbage collector removes all unreachable provenance values from the zero-count-tables, but this happens before it determines if these unreachable provenance values actually correspond to nodes in the tree that we can remove.

PR #258, opened 2026-06-30
Allocate slots for protected tags from a dedicated region.

icmccorm:
We are overallocating space for protected tags. The following assertion will periodically fail for concrete tags passed to `__bsan_protector_end`:
```rust
impl ProtectedTags {
    pub fn remove_protector(&mut self, tag: BorTag) {
        assert!(self.0.remove(&tag).is_some());
    }
}
```
This is because of an incorrect assumption that I made about the shadow stack. I thought that function entry retags would always dominate other operations that use the shadow stack. This is not the case; there can be load instructions before function entry retags. 

This PR improves documentation surrounding the `ShadowStackAllocator`, and it reserves space for the maximum number of possible function entry retags up-front. This will affect the efficiency of garbage collection — we may read stale tags — but it is unavoidable.

### DIFF
--- a/bsan-pass/BorrowSanitizer.cpp
+++ b/bsan-pass/BorrowSanitizer.cpp
@@ -769,51 +769,27 @@ PreservedAnalyses BorrowSanitizerPass::run(Module &M,
 }
 
 namespace {
-// BorrowSanitizer uses a shadow stack to track the provenance values
-// that are accessible in memory and to pass provenance between functions.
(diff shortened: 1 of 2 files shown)


icmccorm:
I'm not 100% certain it is unavoidable. But my sense is that there are programs for which it is impossible, statically or dynamically, to partition the shadow frame for a function into two distinct regions of protected provenance values and unprotected provenance values without leaving at least some potentially uninitialized gaps in each section.

PR #267, opened 2026-07-02
Retain unprunable dead tags for future GC passes

### LINKED ISSUE #265: Ensure that nodes with multiple children are still pruned, eventually.
When we attempt to prune a node from the tree, and it has multiple "live" children, we [leave it alone](https://github.com/BorrowSanitizer/bsan/blob/032a4fca7039e67db700a1bc3e9d7d1330517fae/bsan-rt/src/tree_borrows/tree.rs#L683).

```Rust
 // Node has exactly one child
if let Some(child_idx) = self.can_be_replaced_by_single_child_dead(idx) {
    // Replace a node with its only child, if it is a root then we promote the child
    // to a root
    ...
}
// Otherwise, the dead node has more than one reachable child and we retain it
// To-do: Consider cases where there are multiple descendants of a parent
// node with varying permissions
```
However, our garbage collector removes dead nodes from the zero-count-tables of each thread. If we cannot prune a node, then it will be leaked: we will never be able to prune it in the future. 

To fix this, we need a way for the Rust side to communicate to the LLVM side about which nodes need to be retained after we attempt to prune a tree. The easiest solution is for the Rust side to receive a mutable slice:
```rust
fn remove_useless_children_dead(&mut self, dead_tags: &mut [BorTag]) { .. }
```
If we remove a node, we will zero-out the corresponding entry in the list of dead tags.

Then, on the C++ side, we will need to check to see which nodes are zeroed-out after we call `__bsan_prune`. Here's [what currently happens](https://github.com/BorrowSanitizer/bsan/blob/bfba71ee673f1bf71ae7e99acc2034f9c709b6e6/bsan-rt/llvm-wrapper/bsan_global.cpp#L88), and where we will need to adjust the current behavior:
```C++
pending_.drain([&](AllocInfo *info, const BorTagSet &tags) {
  // If `__bsan_prune` returns true, then every single tag was pruned.
  if (__bsan_prune(info, tags.data(), tags.size())) {
    if (snap.min_drained == snap.gen) {
      __bsan_eject(info);
    } else {
      quarantine_.push_back({info, snap.gen});
    }
  }else{
  // Nothing happens here yet!
  }
});
```
In the `else` branch, we want to scan the `tags` array, and see which tags are still nonzero. These are the ones that we want to keep around for future collection runs.

Prior to draining the pending set, we want to create a new, empty pending set. As we drain the original set, when we see that a tag needs to be kept around, we will add it to the new pending set. After draining the old set, we will replace it with the new one. This makes it so that the pending set is actually nonempty between GC runs. It will hold tags that we could not remove right away.

These changes should be done with a PR against the GC branch (see #252).

vnt1c:
Closes #265. A dead node with multiple live children cannot be pruned. However, we must keep these nodes so that they can be collected in a future GC pass. I implemented a `pending` set which keeps track of these unpruned nodes so that they can be collected in future GC passes. Primarily,

- `remove_dead_tags` and the associated pruning methods now take &mut [BorTag] and pruned entries are zeroed out.
- Unpruned entries are kept in the list and swapped with the old dead_tags list.
- Note: uninitialized trees now reports itself as empty when its root is dead (and its reference count is zero), allowing the `AllocInfo` metadata to be reclaimed.


vnt1c:
Wrong branch.

PR #272, opened 2026-07-06
Fixed `byval` and guarded against ABI differences.

icmccorm:
This PR fixes `byval` semantics. We need to copy the shadow from the caller to the callee. The caller will have a type-annotated pointer (e.g. `byval(ty)`). We need to load the provenance of this type from the caller, and then store it onto the shadow stack alongside other arguments' provenance. In the callee, if we see a `byval` argument, then we need to load the provenance values that correspond to the type from the shadow stack, as if a value of that type was passed directly as a parameter. Then, we store each provenance value into the shadow location for the callee's copy of the `byval` argument. For context, see the [langref section](https://llvm.org/docs/LangRef.html#parameter-attributes) on `byval` and other parameter attributes.

We also need to guard against ABI differences when we pass integers that could be disguised as pointers. For example:
```rust
#[repr(C)]
struct Foo {
    ptr: *mut u32
}

extern "C" {
   unsafe fn as_value(x: Foo);
}
```
The function `as_value` will pass an instance of `Foo` as an `i64`. However, `clang` will compile the definition of `as_value` to receive a `ptr`. If this pointer is dereferenced, it will get a stale provenance value, leading to a false positive. If `rustc` emitted the byte type (and we supported it), then `as_value` would likely receive a `b64`, so we would know to pass provenance. As a temporary fix, this PR makes it so that we pass omnivalid provenance for zeroed-out integers that are greater than or equal to the size of a pointer, but only if the callee is defined in another module.

### DIFF
--- a/bsan-pass/BorrowSanitizer.cpp
+++ b/bsan-pass/BorrowSanitizer.cpp
@@ -893,7 +893,7 @@ private:
 // | parameter provenance  |
 // |_______________________|
 // |                       |
-// | static allocas        |
+// | static/byval allocas  |
 // |_______________________|
 // |                       |  <-- FnEntryTop
 // | function entry retags |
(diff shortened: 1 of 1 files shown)


mojeanmac:
LG!

PR #273, opened 2026-07-08
One-word provenance part #1

### LINKED ISSUE #266: One word provenance - Part #1
At the moment, we use indices to link `Node` objects to their parent and children within the `Tree`. We need to replace this with direct pointers. The first step to doing this is to replace the current structure, which looks like this:
```rust
struct Tree {
    ...
    pub(crate) tag_mapping: UniKeyMap<BorTag>,
    /// All nodes of this tree.
    pub(super) nodes: UniValMap<Node>,
}
```
Instead, we want a single mapping from `BorTag` instances to `Node` pointers:
```rust
struct Tree {
    ...
    pub(crate) node_mapping: HashMap<BorTag, NonNull<Node>>,
}
```
Before fixing any errors, we also want to change the `Node` object. At the moment, it uses `UniIdx` instances to refer to children and its parent. Change these to `NonNull<Node>`, like so:
```rust
pub struct Node {
    ...
    /// All nodes except the root have a parent node.
    pub parent: Option<NonNull<Node>>,
    pub children: SmallVec<[NonNull<Node>; 4]>,
```
Then, fix the APIs to adjust for this change. For now, you can allocate new `Node` instances using `Box::into_raw` / `Box::from_raw`. These will be replaced with a different allocator. This should be done on a dedicated branch: `main/owp`. Create this branch, and then make a PR into it with these changes, once all tests are passing.

Gitter499:
Closes #266


bsan-rt/src/tree_borrows/data_structures/mod.rs
```diff
@@ -3,4 +3,4 @@ mod range_object_map;
 mod unimap;
 pub use dedup_range_map::*;
 pub use range_object_map::*;
-pub use unimap::*;
+// pub use unimap::*;
```
icmccorm: No need to leave the comment.

bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -270,28 +271,59 @@ pub struct LocationTree {
...
     /// NOTE: not all tags registered in `Tree::nodes` are necessarily in all
     /// ranges of `perms`, because `perms` is in part lazily initialized.
-    /// Just because `nodes.get(key)` is `Some(_)` does not mean you can safely
-    /// `unwrap` any `perm.get(key)`.
+    /// Just because `nodes.contains_key(&tag)` does not mean you can safely
+    /// `unwrap` any `perm.get(&key)`.
     ///
     /// We do uphold the fact that `keys(perms)` is a subset of `keys(nodes)`
-    pub perms: UniValMap<LocationState>,
+    pub perms: FxHashMap<BorTag, LocationState>,
```
icmccorm: This should be a mapping from `NonNull<Node>` to `LocationState`.

bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -270,28 +271,59 @@ pub struct LocationTree {
...
     /// Caches information about the relatedness of nodes for a wildcard access.
     pub exposed_cache: ExposedCache,
 }
+
+pub type NodeMap = FxHashMap<BorTag, NonNull<Node>>;
```
icmccorm: Instead, you should have:

```rust
struct NodeMap(FxHashMap<BorTag, NonNull<Node>>);
```

The `impl` will declare the methods below.

icmccorm:
@Gitter499 I left a few comments. We'll want to keep this on the branch you made until we complete the entire set of changes, so this won't be merged yet. You can start a new PR into the branch `bsan-main/owp` for the next part.

icmccorm:
Closing, part 2 will be a PR into the branch `bsan-main/owp`.

PR #276, opened 2026-07-09
Benchmark BorrowSanitizer under Miri's configuration

icmccorm:
Our benchmarking script was executing tests natively that were ignored by Miri, via `cfg` directives. This lead to us inadvertently comparing BorrowSanitizer's performance running a test case against Miri's performance doing nothing, skewing our results.

This PR skips all tests that are ignored by Miri by parsing Rust's test harness output (`--format=json`). Additionally, it compiles everything using `--cfg=miri`. It also fixes a bug where the raw `.csv` result output recorded the incorrect mean value. This did not affect the `.json` output displayed in our benchmarking dashboard.


mojeanmac:
LGTM

PR #280, opened 2026-07-13
Update to 1.99.0

icmccorm:
Updates our fork of Rust to 1.99.0.


icmccorm:
@mojeanmac - On aarch64, standard library allocator shims are inconsistently given line numbers by the symbolizer. Locally on my Mac, I see `LL:CC` in normalized output for calls to `__rdl_alloc` and the like, but this is missing in the output seen by the arm actions runners. To solve this, I adjusted the UI test script's regexes to remove line number annotations for functions with the `__rdl_` prefix.

Also, this is likely the last time we will update the Rust fork. Soon, we'll be able to switch over to using nightly!

PR #286, opened 2026-07-15
exclude test_bit_vec_iterator in crates.json

mojeanmac:


icmccorm:
Do we know why this was failing?

mojeanmac:
> Do we know why this was failing?

One test segfaults in CI and on my machine, but not on ACEs

Issue #279, opened 2026-07-12
One Word Provenance (part #2)


icmccorm:
We refactored the `Tree` and `Node` objects to use pointers to nodes instead of the `UniIndex` type.
Now, we need to switch over to using `Node` as the method for identifying and validating accesses to an allocation.

Assume that each provenance value is a pointer to an `AllocInfo` object. We want to use the global `Heap` to hand out objects of this kind. The object will have three possible states, which should be modeled as an `enum`:
* `Uninit` - stores the information that we need to create metadata for the allocation. Similar to the `LazyTree` state.
* `Tree` - an initialized `Tree`, containing the root metadata for the allocation.
* `Node` - stores the information for a single `Node`. This should contain a pointer back to the `Tree` (Something along the lines of `Option<NonNull<Tree>>`).

When we first retag an `Uninit` object, we need to initialize it as a `Tree` and then hand out a new `Node` as the result of the retag. Retagging a `Node` produces another `Node`. All read, write, and retag accesses using a `Node` need to lock the `Tree`. However, reference count updates can happen at the level of each `Node`, using atomic semantics. 

As a result of this implementation, the `BorTag` component of a provenance value should be effectively unused.

@Gitter499 This should be a PR from a new branch, into the destination branch `bsan-main/owp`.

icmccorm:
```Rust
// The state of an allocation. This is essentially
// what we currently have for `LazyTree`.
enum BorrowState {
    Init { tree: Tree },
    Uninit { size: usize, offset: usize },
    None,
}

struct Node {
    refcount: RefCount;
    state: NonNull<Mutex<BorrowState>>>;
    // ... all of the remaining fields of Node,
    // 
}

// When we allocate the "root" node of an allocation,
// (e.g. `__bsan_alloc` and `__bsan_alloc_stack``), we
// allocate an instance of `RootNode`, and then return
// a pointer to `Node`. This prevents us from having
// to allocate twice for each root, and it also improves
// the locality of accesses during initialization.
struct RootNode {
    node: Node,
    // the interior pointer within node points
    // to the second field of this struct.
    // we will initialize BorrowState so that it
    // contains BorrowState::Uninit at first.
    state: Mutex<BorrowState>,
}

type Provenance = *mut Node;
```

PR #287, opened 2026-07-16
refactor(#279): rename LazyTree to BorrowState on AllocInfo

Gitter499:
## Summary
- Rename `LazyTree`/`Init` to `BorrowState`/`Tree` and store it as `AllocInfo.borrow`, so allocation borrow state has a named home for one-word provenance work.
- Behavior and ABI are unchanged: provenance is still `(BorTag, AllocInfo*)`; AllocInfo remains the only Heap freelist object.
- Does **not** implement [#279](https://github.com/BorrowSanitizer/bsan/issues/279) (Heap `Uninit|Tree|Node`, one-word provenance, unused tag). That is the follow-up on top of this.

## Test plan
- [x] `cargo +bsan test -p bsan_rt --lib` (43 passed)
- [x] `./xb --skip ui --keep-sysroot` (exit 0)
- [ ] Optional: hashbrown-only bench vs ~484× baseline (expect no change)

### DIFF
--- a/bsan-rt/src/borrow_state.rs
+++ b/bsan-rt/src/borrow_state.rs
@@ -1,0 +1,15 @@
+//! Allocation-level borrow metadata attached to [`crate::AllocInfo`].
+//!
+//! Today this is the old `LazyTree` path under a clearer name: one
+//! [`BorrowState`] per allocation, stored under `AllocInfo`'s mutex.
+//! Provenance remains two-word (`BorTag` + `AllocInfo*`).
+//!
+//! - [`BorrowState::Uninit`]: lazy; root tag / size / span / refcount only.
+//! - [`BorrowState::Tree`]: owns the full [`crate::tree_borrows::tree::EagerTree`].
+//!
+//! [#279](https://github.com/BorrowSanitizer/bsan/issues/279) (one-word provenance)
+//! will turn Heap-allocated metadata into `Uninit | Tree | Node`, make provenance a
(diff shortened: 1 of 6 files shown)


bsan-rt/src/borrow_state.rs
```diff
@@ -0,0 +1,15 @@
+//! Allocation-level borrow metadata attached to [`crate::AllocInfo`].
```
icmccorm, line 1: We don't need this extra file or docs. `BorrowState` should just be a public export of the `tree_borrows` modules

bsan-rt/src/lib.rs
```diff
@@ -237,13 +240,14 @@ impl Debug for FreeListOrAddr {
...
-/// the tree for the allocation.
+/// the [`BorrowState`] attached to this allocation.
 #[repr(C)]
 pub struct AllocInfo {
     alloc_id: Cell<AllocId>,
     free_or_addr: Cell<FreeListOrAddr>,
     size: Cell<Size>,
-    tree: Mutex<Option<AllocStateImpl>>,
+    /// Polymorphic borrow metadata for this allocation (`Uninit` | `Tree`).
+    borrow: Mutex<Option<AllocStateImpl>>,
```
icmccorm, line 250: We want to shift the members of `AllocInfo` into the `BorrowState` object. Where we have `*mut AllocInfo` now, we want `*mut BorrowState`.

icmccorm (approved):

PR #289, opened 2026-07-17
Intercept `clang` invocations instead of setting `CFLAGS` globally.

icmccorm:
When we build [`fiona-rs`](https://github.com/cmazakas/fiona-rs) using `cargo-bsan`, it triggers a linker error. Build scripts for C dependencies might create small binaries to test for certain platform-specific features or behaviors. These binaries were being instrumented by BorrowSanitizer, but our runtime was never linked, which triggered missing symbol errors. This happens because we do not include the flags for linking our runtime as part of `CFLAGS`. Instead, we wait until we build the final, combined Rust/C/C++ binary. 

This PR fixes this error through a combination of features. 

First, `cargo-bsan` will now redirect all `clang` invocations to itself via a symlink (`./target/bsan/clang`). Then, we manually pass a variety of custom `CFLAGS` to `clang`, along with the arguments specified in the original invocation. Our custom arguments instruct `clang` to run our instrumentation pass, link our runtime, and also link the core system dependencies required by the sanitizer interceptors. When you use one of the upstream sanitizers, `clang` will add these [by default](https://github.com/llvm/llvm-project/blob/f010423d3dfac774c6157686b6f14fb8d357c715/clang/lib/Driver/ToolChains/CommonArgs.cpp#L1590).

This PR also simplifies how we link our runtime. Originally we included the entire static archive by passing `--whole archive` to the linker. This was necessary because the LLVM wrapper only uses the Rust core via weak symbols. The linker would discard the Rust runtime, because strong symbols are necessary to pull dependencies from a static archive. 

Instead, we define a custom, unused "anchor" symbol:

```C++
extern "C" void __bsan_rust_runtime_anchor(void);
USED static void (*const bsan_rust_runtime_anchor)(void) =
    &__bsan_rust_runtime_anchor;
```
This forces the linker to include both the Rust and C++ components. We also need to manually define this symbol when running in no-op mode, so that the linker doesn't think that it is missing. 

Finally, when we build the Rust component of our runtime, we need to pass `-Zbuild-std=core,alloc` to `cargo`, so that we propagate `-Cpanic=abort` throughout all of its dependencies. Otherwise, we will be left with undefined `rust_eh_personality` symbols when we attempt to link our runtime to a program without any Rust dependencies. 

Combined, these changes make it possible to build and test `fiona-rs` simply via:
```
cargo +bsan bsan test
```

### DIFF
--- a/cargo-bsan/src/phases.rs
+++ b/cargo-bsan/src/phases.rs
@@ -233,7 +233,7 @@ pub fn phase_cc(args: impl Iterator<Item = String>) {
     // The sanitizer runtime requires a set of default system libraries
     // to always be linked. Here, we pass them with `--no-as-needed` to
     // ensure that these libraries are never excludeddue to other linker
-    // configuration.
+    // configurations.
     //
     // (see llvm-project/clang/lib/Driver/ToolChains/CommonArgs.cpp#L1590)
     cmd.arg("-Wl,--push-state,--no-as-needed");
--- a/cargo-bsan/src/phases.rs
+++ b/cargo-bsan/src/phases.rs
@@ -1,4 +1,5 @@
 use std::path::PathBuf;
+use std::process::Command;
 
 use rustc_version::VersionMeta;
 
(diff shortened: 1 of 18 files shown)


cargo-bsan/src/phases.rs
```diff
@@ -201,15 +209,47 @@ pub fn phase_cargo_bsan(mut args: impl Iterator<Item = String>) {
...
+        // as a linker, which is flaky, we allow warnings.
+        if arg == "-Werror" {
+            continue;
+        }
+        cmd.arg(arg);
+    }
+
+    // The sanitizer runtime requires a set of default system libraries
+    // to always be linked. Here, we pass them with `--no-as-needed` to
+    // ensure that these libraries are never excludeddue to other linker
```
icmccorm:
```suggestion
    // ensure that these libraries are never excluded due to other linker
```

cargo-bsan/src/phases.rs
```diff
@@ -201,15 +209,47 @@ pub fn phase_cargo_bsan(mut args: impl Iterator<Item = String>) {
...
+    // configuration.
```
icmccorm:
```suggestion
    // configurations.
```

PR #292, opened 2026-07-22
Match back-edge behavior between `remove_useless_children` and `remove_useless_children_dead` (OWP#1)

Gitter499:
In `remove_useless_children_dead`, single-child compact now sets `child.parent` before rewriting the parent's children/roots slot, matching `remove_useless_children`.


mojeanmac:
Can you merge main into this? We don't replace root nodes anymore, only removing them when they're singletons, so the code looks different now (this may not be relevant anymore)

PR #293, opened 2026-07-22
One-word Provenance (OWP#2)

Gitter499:
## WIP 🚧

- Provenance is `*mut Node` (sentinels 0/1/2); RC is atomic on `Node`.
- Pass/shadow store one 8-byte word; former ORIGIN VA ranges are INVALID.
- Shadow Copy/Move/Clear/Write operate on one `Node*` slot; RC uses inc-before-dec (#283).
- Null pointer constants get invalid provenance (#291).

```rust
pub(crate) struct RootNode {
    pub(crate) node: Node,
    pub(crate) alloc_id: Cell<AllocId>,
    pub(crate) base_addr: Cell<Size>,
    pub(crate) size: Cell<Size>,
    pub(crate) state: Mutex<BorrowState>,
    pub(crate) free_list_next: Option<NonNull<RootNode>>,
    pub(crate) on_stack: bool,
}

pub enum BorrowState {
    None,
    Uninit { root_tag: BorTag, size: Size, span: Span },
    Init(EagerTree),
}
```

### DIFF
--- a/bsan-pass/BorrowSanitizer.cpp
+++ b/bsan-pass/BorrowSanitizer.cpp
@@ -37,9 +37,9 @@
 using namespace llvm;
 using namespace llvm::PatternMatch;
 
-// Provenance is two words: a borrow tag and
-// a pointer to an allocation metadata object.
-static const unsigned kProvenanceSize = 16;
+// Provenance is one word: a Node* (sentinels 0/1/2 for
+// omnivalid/invalid/wildcard).
+static const unsigned kProvenanceSize = 8;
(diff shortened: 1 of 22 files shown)


icmccorm:
What is on_stack used for?

Gitter499:
It's used as a guard in `prune_dead_tags` and `bsan_eject` so that we don't free the `RootNode` while it is owned by the stack frame.

It is a redundant debugging artifact since we already take into account all the live provenance values in our GC implementation.

PR #294, opened 2026-07-22
Key GC cleanup by root Node* (OWP#3)

Gitter499:
Shadow/RC still pass child `Node*`; ZCT/live/pending keys canonicalize via `__bsan_alloc_root`

hashbrown@0.17.0 (aarch64-unknown-linux-gnu), 103 tests, mean relative execution time:

| Mode | vs native | vs miri-tb |
|------|----------:|-----------:|
| full | 654× | 0.246× |
| no-op | 7.50× | 0.0029× |

### DIFF
--- a/bsan-rt/llvm-wrapper/bsan_global.cpp
+++ b/bsan-rt/llvm-wrapper/bsan_global.cpp
@@ -33,7 +33,7 @@ void GlobalContext::MergeZeroCounts(const uptr, BsanThread *const &thread,
   // count table, then we can drain its contents for garbage collection.
   if (!thread->ZctBusy()) {
     // Move unreachable tags into pending; keep live ones in this thread's ZCT.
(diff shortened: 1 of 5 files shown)


icmccorm:
How does this compare to hashbrown prior to the changes?

Gitter499:
Ran both on my local machine and it is about twice as slow (on full) as our two-word provenance but faster in no-op mode. I've only tested it against `hashbrown`, which is pretty retag heavy. I will run the full benchmarks later tonight to see what the effects are on a more broad crate corpus


Metric | main | OWP
-- | -- | --
full / native | 368× | 654×
full / miri-tb | 0.143× | 0.246×
no-op / native | 8.15× | 7.50×
no-op / miri-tb | 0.0034× | 0.0029×

Gitter499:
Though note that these are early numbers! I am working on some optimizations that might lower the overhead, but we will see

PR #288, opened 2026-07-16
Multi-Child Compaction & GC Minimum Tree Size

vnt1c:
Relevant to bullets 1 and 2 of #275. Implemented multi-child compaction and a minimum tree size guard before pruning trees. Rather than retaining dead nodes with multiple children, the GC can now safely remove these nodes under [*some*](https://github.com/BorrowSanitizer/bsan/pull/288/changes#diff-c7b5f33c39d5433305ba319a5a0cd2d20e76fa4007c0add46beaefdd6ff9e90dR431) conditions. This reduces the accumulation of dead interior nodes, which is especially impactful in bigger trees. Multi-child compaction yielded a **mean speed up of 4.75%** (measured by execution time of test suites across 106 Rust crates). 

While the minimum tree size garbage collection guard has not been profiled, it can be logically argued that this can only lead to performance benefits (assuming the cost of traversing through dead nodes of smaller trees is cheaper than pruning those nodes). A simple local Hyperfine test of this change across the three crates in the repository showed faster performance in all three cases. Future work can change the minimum tree size (it is currently set at 64) or attempt an empirical approach (a formula) to dynamically determine this count at runtime.

The results for execution times for Rust crate tests before and after multi-child compaction can be found below:
| Crate | # of Tests | Mean Before | Mean After | Δ | % Change |
|---|---|---|---|---|---|
| memchr-2.8.1 | 1 | 1056.438s | 881.058s | -175.380s | -16.601% |
| crc32fast-1.5.0 | 6 | 59.270s | 54.499s | -4.771s | -8.049% |
| encoding_rs-0.8.35 | 15 | 55.048s | 53.693s | -1.356s | -2.463% |
| concurrent-queue-2.5.0 | 10 | 47.925s | 50.547s | +2.622s | +5.470% |
| minimal-lexical-0.2.1 | 26 | 25.346s | 33.383s | +8.036s | +31.707% |
| base64ct-1.8.3 | 27 | 16.562s | 18.840s | +2.278s | +13.751% |
| unicode-normalization-0.1.25 | 13 | 15.771s | 15.544s | -0.227s | -1.442% |
| matchers-0.2.0 | 17 | 14.375s | 14.438s | +0.062s | +0.434% |
| bit-vec-0.9.1 | 40 | 12.513s | 10.872s | -1.641s | -13.113% |
| parking_lot_core-0.9.12 | 1 | 11.813s | 12.134s | +0.320s | +2.712% |
[...]

PR #295, opened 2026-07-23
GC nits

### LINKED ISSUE #275: GC Optimizations
Four techniques for optimizing Miri's tracing GC (on [this branch](https://github.com/cmu-cornell-for-rust/miri/tree/lazy%2Bgc))(Also may be easier to view the diff [here](https://github.com/cmu-cornell-for-rust/miri/pull/1/changes)) that may still be usefull in a deferred reference counting context for BorrowSanitizer:

## 1. Minimum tree size for garbage collection

- [x] This is probably the simplest optimization to be made, but it resulted in decent average speedups in Miri. I found that only GCing trees with <= 64 nodes worked in Miri. I suspect the best size will be similar in BorrowSanitizer because we are not fundamentally changing TreeBorrows that much.

## 2. Multi-child compaction

- [x] Cleaning up a node with multiple children is sound if *each* child node still obeys `perms::can_be_replaced_by_child` AND both parent and any child are not `ReservedIM`. This has already been implemented in the branch of Miri linked above, entirely in the TreeBorrows module, so it should be copy-pasteable to transfer over the new ruleset, the updated method in `tree`, and the new test in `tree/tests.rs`.

## 3. Visit-based activation

- [x] Rather than base the interval for GC activation on retags, I found that a visit-count is a better metric to see how much trees are *in use* rather than how big they've grown. ~20k visits works for Miri, but our GC may be more/less intense, so its likely we'll need a new constant.

Inside `tree::perform_normal_access` and `perform_wildcard_access`, there's a there are closures assigned to `node_app` where every iteration corresponds to visiting a node. Be careful of what you put inside this closure, as instructions in here will be iterated millions of times over the course of execution! The current activation metric in BorrowSanitizer is `MaybeRequestGC()` in `bsan.cpp` on the C++ side.

## ~~4. Minimum tree size to count towards activation~~

Edit: Ultimately these nodes and visits contribute so little to the total visit count, that it's not worth the extra checks during each access.
- ~~My intuition is that we also wouldn't want visits from smaller trees to impact the visit count, but I'm still working on having done it successfully. Like above, there are costs to having checks every visit, so avoid doing that. I'd expect we'd want to skip counting visits for trees smaller than 64 nodes?~~

## Testing on ACES

I recommend creating `BSAN_OPTION` flags for each of the constants in the above techniques, so that you can set them before running without having to rebuild the whole image.

You can build bsan containers on aces with the build script in `/containers/build_scripts` just run `./build.sh bsan <commit hash or branch name>` to build a container named `bsan-<commit or branch>`. Without a second arg, it will build from main.

In `/scripts` I suggest running `./run_bsan_dataset <container> 2:00 top_500_fast <optional BSAN_OPTIONS>`. Your result should show up in `/outputs` in a csv with the container name and any options added.

mojeanmac:
This PR makes minor changes to (mainly `tree.rs`) from #288. There were some regressions in our benchmarks, but it's more of a sign that our set of benchmarks is well rounded :)

To attempt to recover these runtimes, I made a few changes, namely to avoid extra checks and use more built in functional language features (under the assumption they are optimized more when compiled down):

- `can_be_replaced_by_single_child` now returns a boolean, since we can safely assume the child's `UniIndex` is at `node.children[0]`, and it now matches the return type of `can_be_replaced_by_children`
- Switches from if-else branching in `remove_useless_children` to a match statement (also more readable)
- Switches to functional-style iterators over nested loops (somewhat of a style nit but Miri also does this)
- Moves root checking inside the `tree::can_be_replaced_by...` functions and removes extra length checks

And quality of life fixes:
- Removes the `GC_MIN_TREE_SIZE` constant - it was previously assigned in the middle of `tree.rs`. This way we're semantically constant between flags/variable names with `tree_gc_min_size`
- Reorders some of `tree.rs` so that all definitions (and small impls if available) are at the start rather than at the end

By adding `visits_per_gc=10,tree_gc_min_nodes=0` to `ui.rs`, I saw the e2e runtime go from 57.8s -> 57.3s (-0.5s) between this branch and main, which hopefully is a good sign, considering that the GC should not be doing much work each pass in our small test cases.

Finally closes #275

### DIFF
--- a/bsan-rt/src/tree_borrows/tree.rs
+++ b/bsan-rt/src/tree_borrows/tree.rs
@@ -1,4 +1,6 @@
-// Ported from Miri (commit:072a9fa) with edits: removed ProvenanceExtra & VisitProvenance
+// Ported from Miri (commit:072a9fa) with edits: removed `ProvenanceExtra` & `VisitProvenance`
+// - added `EagerTree` and `LazyTree`
+// - adapted GC functions for shadow memory reference counter
 //! In this file we handle the "Tree" part of Tree Borrows, i.e. all tree
 //! traversal functions, optimizations to trim branches, and keeping track of
 //! the relative position of the access to each node being updated. This of course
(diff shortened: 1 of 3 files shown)


bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -438,80 +494,59 @@ impl EagerTree {
...
 
     /// Like [`Self::can_be_replaced_by_single_child`], but for a node with more than one
     /// child. This requires the stronger [`Permission::can_be_replaced_by_children`] check, and
     /// it must hold for every child at every location
     fn can_be_replaced_by_children(&self, idx: UniIndex) -> bool {
         let node = self.nodes.get(idx).unwrap();
-
-        // The caller only reaches this branch after `can_be_replaced_by_single_child`
-        // has failed, so we do not need a redundant multi-child check
-        if node.children.len() <= 1 {
```
vnt1c: Since the caller already failed the `can_be_replaced_by_single_child` check in the `remove_useless_children` function, I put this guard here so that it doesn't run the same check again 

i.e., if a node fails the single child replacement test, it would also fail the multi-child compaction test

vnt1c (approved):
LGTM, this is much cleaner!

PR #297, opened 2026-07-28
Disable stack instrumentation via a flag

Gitter499:
## Results

Running `hashbrown` in full mode:  320x slower than native and 7.6x faster than Miri. No-op: 6.7 slower than native. 
Interestingly, the range was much lower across tests.

Pass `--no-stack` to the benchmarking script to run without instrumenting stack 

## Precision

Disabling stack instrumentation still allowed us to report the same UB error as found in https://github.com/cmazakas/fiona-rs/issues/5


bsan-pass/BorrowSanitizer.cpp
```diff
@@ -48,6 +48,22 @@ static cl::opt<bool> ClHandleAsmConservative(
...
+    cl::desc("Disable instrumentation of stack allocations (allocas and byval "
+             "arguments). Accesses through pointers to the stack are never "
+             "checked."),
+    cl::Hidden, cl::init(false));
+
+// rustc parses `-Cllvm-args` before loading plugins passed via
+// `-Zllvm-plugins`, so our options are not registered yet at parse time. The
+// environment provides an alternative channel for setting this option when
+// the pass is loaded through rustc instead of `opt`.
+static bool disableStackInstrumentation() {
```
icmccorm, line 68: Can we turn this into a function or macro, to make it easier to specify other environment flags in this way?

bsan-pass/BorrowSanitizer.cpp
```diff
@@ -48,6 +48,22 @@ static cl::opt<bool> ClHandleAsmConservative(
              "outputs to wildcard Provenance"),
     cl::Hidden, cl::init(true));
 
+static cl::opt<bool> ClDisableStackInstrumentation(
+    "bsan-disable-stack-instrumentation",
```
icmccorm, line 52: Let's disable instrumenting `byval` allocas as a separate option from `allocas`. By default, we should skip `byval` allocas, since they are compiler-inserted and unlikely to be involved in aliasing violations. We don't need an environment flag for this additional option.

icmccorm (changes requested):

bsan-script/etc/bench.py
```diff
@@ -56,6 +56,14 @@
 # Every Miri configuration.
 MIRI_CONFIGS = [MIRI]
 
+# Full checking, with stack instrumentation disabled.
+# Added to the set of configurations by the `--no-stack` flag.
+NO_STACK_CONFIG = {
```
icmccorm: We should add this as another `BSAN_CONFIGS` entry, so that it gets run automatically during benchmarking.

bsan-pass/BorrowSanitizer.cpp
```diff
@@ -58,6 +58,42 @@ static cl::opt<bool> ClHandleAsmConservative(
...
```
icmccorm, line 62: This should be `bsan-inst-allocas` (set to true by default) and then `bsan-inst-byval` (set to *true* by default).

I just realized, if someone creates a pointer `&raw mut x`, where `x` is a `byval` argument, then this will likely make the implicit allocation directly accessible to the user, making UB possible. Instead, I think we should instrument these by default.

bsan-pass/BorrowSanitizer.cpp
```diff
@@ -58,6 +58,42 @@ static cl::opt<bool> ClHandleAsmConservative(
...
+}
+
+bool llvm::disableStackInstrumentation() {
+  return flagEnabled(ClDisableStackInstrumentation,
+                     "BSAN_DISABLE_STACK_INSTRUMENTATION");
+}
+
+// `byval` allocations are stack allocations, so disabling stack
+// instrumentation overrides the option to instrument them.
+static bool instrumentByValArgs() {
```
icmccorm, line 93: I don't think that this needs to be a separate function.

bsan-pass/BorrowSanitizerPass.h
```diff
@@ -6,6 +6,11 @@
 
 namespace llvm {
 
+// Defined in BorrowSanitizer.cpp. True if stack instrumentation is disabled
```
icmccorm, line 9: Likewise, I don't think we need this declaration.

bsan-pass/BorrowSanitizer.cpp
```diff
@@ -1741,15 +1777,22 @@ class BorrowSanitizerVisitor : public InstVisitor<BorrowSanitizerVisitor> {
...
         ByValArgInfo Info;
         Info.Arg = &Arg;
-        Info.AllocProv = Prov;
         Info.Size = Size;
 
+        // By default, the implicit stack allocation backing a byval argument
+        // is not instrumented: we never create allocation metadata for the
+        // callee's private copy, so accesses through it are not tracked as a
+        // stack allocation. Field provenance is still copied below so
+        // pointers passed by value keep their provenance.
```
icmccorm, line 1788: This comment will need to be adjusted to reflect that `byval` allocas are now instrumented by default. I don't think we need to justify this in terms of Rust semantics in the documentation, though.

PR #301, opened 2026-07-29
Add more benchmarks

Gitter499:
Follow up to #299 to add more benches to our benchmarking dashboard!

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -2,7 +2,7 @@ name: Benchmarks
 on:
   push:
     branches: [ "main" ]
-  workflow_dispatch: 
+  workflow_dispatch:
   schedule:
     - cron: '0 0 * * *'
 permissions:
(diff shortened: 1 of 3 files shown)


icmccorm:
Let's remove aho-corasick for now. @mojeanmac, are there any other crates that should be included here?

mojeanmac:
Yeah! I have a list of crate testbenches that run in reasonable time (<2hrs in Miri), with the tests that should be ignored [here](https://github.com/cmu-cornell-for-rust/ACES-backup/blob/f15034c3cc0444bdfdc0994cca264d925ba4a99f/outputs/ignore_list-top_500_fast.csv). Maybe you could transpose this into the json format our benchmarks script takes? I'd still ignore that `bit-vec` test which isn't explicitly ignored by Miri but causes us to run OOM for now.

Gitter499:
Replaced with `thread_local`! Waiting for benchmark job results

PR #308, opened 2026-08-09
Remove protection from inlined retags.

### LINKED ISSUE #306: False positive: recursive `&mut self` reborrow, then accesses after the call returns
**Version**: BorrowSanitizer 0.1.0 (46d4afbe4e 2026-08-07), rustc 1.99.0-dev, `bsan` toolchain from `ghcr.io/borrowsanitizer/bsan:latest`, aarch64 Linux

## Reproduce

Run the following code with `bsan` in **release** mode (debug mode is ok).

```rust
use std::marker::PhantomData;

#[derive(Debug, Clone, Copy, PartialEq)]
enum Encoding {
    Module,
    Component,
}

#[derive(Debug, Clone, Copy, PartialEq)]
enum State {
    Header,
    SectionStart,
    FunctionBody,
}

struct Reader<'a> {
    data: &'a [u8],
    pos: usize,
    original_offset: usize,
}

impl<'a> Reader<'a> {
    fn new(data: &'a [u8], original_offset: usize) -> Reader<'a> {
        Reader { data, pos: 0, original_offset }
    }
    fn pos(&self) -> usize {
        self.original_offset + self.pos
    }
    fn read_u8(&mut self) -> Result<u8, ()> {
        if self.pos < self.data.len() {
            let b = self.data[self.pos];
            self.pos += 1;
            Ok(b)
        } else {
            Err(())
        }
    }
}

enum Payload {
    Version { num: u16, encoding: Encoding },
    End,
}

enum Chunk {
    Parsed { consumed: usize, payload: Payload },
}

struct Parser {
    state: State,
    offset: u64,
    encoding: Encoding,
}

impl Parser {
    fn new(offset: u64) -> Parser {
        Parser {
            state: State::Header,
            offset,
            encoding: Encoding::Module,
        }
    }

    fn parse<'a>(&mut self, data: &'a [u8], eof: bool) -> Result<Chunk, ()> {
        let starting_offset = self.offset as usize;
        let mut reader = Reader::new(data, starting_offset);
        match self.parse_reader(&mut reader, eof) {
            Ok(payload) => {
                let consumed = reader.pos() - starting_offset;
                self.offset += consumed as u64;
                Ok(Chunk::Parsed { consumed, payload })
            }
            Err(_) => Err(()),
        }
    }

    fn parse_reader(&mut self, reader: &mut Reader<'_>, eof: bool) -> Result<Payload, ()> {
        match self.state {
            State::Header => {
                let mut header = 0u32;
                for _ in 0..4 {
                    header |= (reader.read_u8()? as u32) << 0;
                }
                self.encoding = if header == 0x6d736100 { Encoding::Module } else { Encoding::Component };
                self.state = State::SectionStart;
                Ok(Payload::Version { num: header as u16, encoding: self.encoding })
            }
            State::SectionStart => {
                if eof && reader.pos() >= 4 {
                    // Simulate: code section present, consume its bytes,
                    // move to FunctionBody so the next call recurses.
                    let _ = reader.read_u8()?;
                    self.state = State::FunctionBody;
                    Ok(Payload::End)
                } else {
                    Ok(Payload::End)
                }
            }
            State::FunctionBody => {
                // KEY: recursively reborrow `&mut self` from inside `&mut self`.
                self.state = State::SectionStart;
                self.parse_reader(reader, eof)
            }
        }
    }
}

fn main() {
    let mut parser = Parser::new(0);
    let data = [0x00u8, 0x61, 0x73, 0x6d, 0x0a, 0x01, 0x00];
    let mut d = &data[..];
    for _ in 0..8 {
        match parser.parse(d, true) {
            Ok(Chunk::Parsed { consumed, payload }) => {
                let _ = payload;
                if consumed == 0 {
                    break;
                }
                d = &d[consumed..];
            }
            Err(_) => break,
        }
    }
    println!("ok");
}
```

The trigger line is the recursive call in `State::FunctionBody`:

```rust
self.state = State::SectionStart;
self.parse_reader(reader, eof)     // recursive &mut self reborrow
```

Report:

```
error: Undefined Behavior: reborrow through <1011>(unprotected) (root of the allocation) at alloc143[0x10] is forbidden
 --> play2.44d5a84660a1493a-cgu.0:0:0
  = help: the accessed tag <1011>(unprotected) (root of the allocation) is foreign to the protected tag <1023>(StrongProtector) (i.e., it is not a child)
  = help: this reborrow (acting as a foreign read access) would cause the protected tag <1023>(StrongProtector) (currently Unique) to become Disabled
  = help: protected tags must never be Disabled
help: the accessed tag <1011>(unprotected) was created here
 --> play2.44d5a84660a1493a-cgu.0:0:0
help: the protected tag <1023>(StrongProtector) was created here, in the initial state Reserved
 --> play2.44d5a84660a1493a-cgu.0:0:0
help: the protected tag <1023>(StrongProtector) later transitioned to Unique due to a child write access at offsets [0x10..0x18]
 --> play2.44d5a84660a1493a-cgu.0:0:0
  = help: this transition corresponds to the first write to a 2-phase borrowed mutable reference

stack backtrace:
0: <play2::Parser>::parse
      at play2.44d5a84660a1493a-cgu.0
1: play2::main
      at play2.44d5a84660a1493a-cgu.0
2: <fn() as core::ops::function::FnOnce<()>>::call_once
      at play2.44d5a84660a1493a-cgu.0

note: stack trace was truncated after 3 frames; set stacktrace_max_len (e.g. BSAN_OPTIONS=stacktrace_max_len=32) to capture more.
```

As evidence of a false positive, the code has no `unsafe` blocks and passes Miri under `-Zmiri-tree-borrows`.

icmccorm:
A tentative fix for #306.

When optimizations are enabled, functions with function-entry retags might be inlined. If a function-entry retag is inlined into the body of a loop, then slots containing protected tags might be overwritten, or the number of function entry retags could be miscounted. This PR avoids the problem by turning function-entry retags into "regular" retags when they occur within a cycle of the CFG. With inlining disabled, function-entry retags will always occur within an acyclic subgraph.

### DIFF
--- a/bsan-pass/BorrowSanitizer.cpp
+++ b/bsan-pass/BorrowSanitizer.cpp
@@ -1284,7 +1284,7 @@ public:
     DTU.flush();
     Cycles.compute(F);
 
-    Plan.build();
+    Plan.build(Cycles);
 
     BasicBlock *EntryBlock = &F.getEntryBlock();
     IRBuilder<> EntryIRB(EntryBlock, EntryBlock->getFirstNonPHIIt());
--- a/bsan-pass/InstrumentationPlan.cpp
+++ b/bsan-pass/InstrumentationPlan.cpp
@@ -58,7 +58,7 @@ void InstrumentationPlan::collectChecks(Instruction *Inst) {
   Checks.emplace_back(Inst, AccessKind, Ptr, AccessRange(DL, AccessSize));
 }
 
-void InstrumentationPlan::build() {
+void InstrumentationPlan::build(CycleInfo &CI) {
   // Collect each of the instructions might propagate provenance metadata.
   for (BasicBlock *BB : depth_first<BasicBlock *>(&F.getEntryBlock())) {
     for (Instruction &I : *BB) {
@@ -77,8 +77,19 @@ void InstrumentationPlan::build() {
       if (auto *CB = dyn_cast<CallBase>(&I)) {
         if (IsRetag(CB)) {
           Retags.push_back(CB);
-          if (IsFnEntryRetag(CB))
-            NumFnEntryRetags += 1;
+
+          RetagInfo RI(CB);
+          if (RI.isProtected()) {
+            // If we see a function-entry retag within a cyclic subgraph
+            // of the CFG, then we know that a function was inlined.
+            // Protectors do not play well with inlining, so we treat this as
+            // a regular retag.
+            if (CI.getCycle(CB->getParent())) {
+              RI.stripProtector();
+            } else {
+              NumFnEntryRetags += 1;
+            }
+          }
         }
         if (auto *LI = dyn_cast<LifetimeIntrinsic>(CB)) {
           AllocaInst *AI = findAllocaForValue(LI->getArgOperand(0), true);
--- a/bsan-pass/InstrumentationPlan.h
+++ b/bsan-pass/InstrumentationPlan.h
@@ -108,7 +108,7 @@ public:
                       const StackSafetyGlobalInfo &SSGI)
       : F(F), DL(DL), DT(DT), SSGI(SSGI) {}
 
-  void build();
+  void build(CycleInfo &CI);
 
 private:
   // The static allocations that we will instrument.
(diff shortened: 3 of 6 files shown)


icmccorm:
This overcorrects a bit. Instead of removing protection entirely, we just need to apply it once.

Issue #315, opened 2026-08-11
False positive: unaligned stores should clear provenance.


icmccorm:
We store provenance in 8-byte aligned chunks. This can lead to false positives for unaligned stores. When a pointer is written to an unaligned address in main memory, its provenance will be written to an aligned address in shadow memory. Subsequent copies can shift a pointer's bits such that they no longer correspond with the correct provenance value in shadow memory. Here's a minimal example:
```Rust
let p_alloc = Box::new(0);
let q_alloc = Box::new(0);

let p = Box::<u8>::as_ptr(&p_alloc);
let q = Box::<u8>::as_ptr(&q_alloc);

// Create source and destination arrays that are four words long.
let mut src = [0u32; 8];
let mut dst = [0u32; 8];

let src_bytes = src.as_mut_ptr().cast::<u8>();
let dst_bytes = dst.as_mut_ptr().cast::<u8>();

unsafe {

    src_bytes.add(8).cast::<*const u8>().write(q);

    //   SRC:
    //        main: |   |   | q | q |   |
    //      shadow: |   |   | q | q |   |

    src_bytes.add(4).cast::<*const u8>().write_unaligned(p);

    //   SRC:
    //        main: |   | p | p | q |   |
    //      shadow: | p | p | q | q |   |

    std::ptr::copy_nonoverlapping(src_bytes, dst_bytes.add(4), 20);


    //   DEST:
    //        main: |   |   | p | p | q |
    //      shadow: | p | p | q | q |   |
    //                        ^   ^

    let p2 = dst_bytes.add(8).cast::<*const u8>().read();
    assert_eq!(p2, p);

    println!("{}", *p2);  // False positive UB! We're accessing p using q's provenance.
```
I'm surprised that we've been able to avoid this error so far. It's likely a combination of us running without optimizations, and unaligned pointer stores being generally uncommon. CHERI has the same problem, but they enforce alignment and error otherwise. We may want to support a mode where we throw an exception on unaligned provenance copies, but by default we should be optimistic and write omnivalid provenance here.

PR #319, opened 2026-08-12
Clear provenance on unaligned memory transfers

icmccorm:
When a the source and destination of a memory transfer (e.g. `memmove`, `memcpy`) have different alignment, then it risks desynchronizing a pointer's address from its provenance in shadow memory (see #315). This PR refactors our helper methods for adjusting shadow alongside these operations, so that when we detect this case, we clear the shadow of the destination instead of transferring the unaligned source. This is not a complete fix yet, but it does allow us to support the minimal example in that issue.


mojeanmac:
LGTM- iirc when we talked about this a few weeks ago, if we drop provenance completely for misaligned memory we would miss true positive cases. Is that the case for this PR?

PR #321, opened 2026-08-13
Ensure that exiting threads queue unpruned nodes for GC

icmccorm:
Every thread maintains a "zero-count-table"; a set of nodes with a zero reference count that have yet to be pruned by the garbage collector. When a thread exits, its ZCT is destroyed, which causes us to leak provenance. This PR adds a global ZCT that is used to store the contents of exiting threads' ZCTs.


mojeanmac (approved):
LGTM! And a global ZCT might be a better GC activation heuristic than node visits.

PR #325, opened 2026-08-18
Store protectors within each `Node`.

icmccorm:
We use a global `RwLock<FxHashMap<..>` to store the protection status of each `Node`. This is a design decision taken from Miri. It's convenient for Miri to have protectors located in a single spot, instead of distributed throughout each tree.

However, we are running in a concurrent environment. Multiple threads will be adding and removing protectors, causing contention on the lock. This PR moves `ProtectorKind` into each `Node` in hopes of reducing this contention.

I've seen a marginal decrease in runtime from this change when comparing individual tests between Miri and BorrowSanitizer for the crate `half`. However, there's a *huge* change when you compare the performance when running the entire test suite.
```
Benchmark 1 (before this PR):
  Time (mean ± σ):      2.454 s ±  0.089 s    [User: 9.716 s, System: 0.926 s]
  Range (min … max):    2.360 s …  2.636 s    10 runs

Benchmark 1 (after this PR):
  Time (mean ± σ):      1.020 s ±  0.011 s    [User: 1.380 s, System: 0.747 s]
  Range (min … max):    1.005 s …  1.038 s    10 runs
```


icmccorm:
Whups, seems like protectors aren't getting removed in the correct location...

PR #344, opened 2026-09-10
Handle deadly signals gracefully.

### LINKED ISSUE #337: Handle SEGV gracefully.
Many of the test cases that remain from Miri involve patterns like this
```Rust
//@run:1
#[allow(deref_nullptr)]
fn main() {
    let x: i32 = unsafe { *std::ptr::null() }; //miri: ~ ERROR: null pointer
    panic!("this should never print: {}", x);
}
```
The pointer created by `ptr::null()` is cast from an integer, so it receives wildcard provenance. However, this provenance doesn't correspond to any exposed allocation, so we optimistically assume that it's valid. The dereference triggers a segfault, and the program terminates.

Other sanitizers handle this kind of error using a signal handler, which will provide additional information when a segfault is triggered. The signal handler is "installed" during the initialization of the runtime (e.g. `__bsan_init`);
```C++
// Here's what happens in MemorySanitizer
InstallDeadlySignalHandlers(MsanOnDeadlySignal);
```
This signal handler will call the following function, which is provided by LLVM's CompilerRT:
```C++
HandleDeadlySignal(siginfo, context, GetTid(), &OnStackUnwind, nullptr);
```
We should install a similar handler for BorrowSanitizer. This will allow us to "bless" the output from nearly all of the tests in `./tests/miri-tests/should-fail`.

vnt1c:
Closes #337. Segmentation faults (along with other faults, as defined [here](https://github.com/llvm/llvm-project/blob/18a61f5a44dda16720754584aee95cac652e3646/compiler-rt/lib/sanitizer_common/sanitizer_posix_libcdep.cpp#L251)) now have signal handlers that are installed at runtime which handle them gracefully. Such faults also now produce a symbolized BorrowSanitizer report and exit (the report format comes from LLVM, which is inconsistent with our usual diagnostic style). 

All five SEGV-class test cases in `tests/miri-tests/should-fail` now pass: `null_pointer_deref`, `null_pointer_write`, `wild_pointer_deref`, `reading_half_a_pointer`, `zst_local_oob`.

Additionally, I've added per-thread alternate stacks for signal handlers, so that we can run signal handlers in cases of stack overflows.


icmccorm:
Can you move the relevant tests into the fail directory and bless the output?

tests/ui.rs
```diff
@@ -277,6 +277,9 @@ regexes! {
     "([0-9]+: .*)::<.*>"             => "$1",
     // erase long hexadecimals
     r"0x[0-9a-fA-F]+[0-9a-fA-F]{2,2}" => "$$HEX",
+    // x86_64 omits the fault address for non-canonical addresses and adds a hint
```
icmccorm, line 280: Is this only on x86?

bsan-rt/llvm-wrapper/bsan.cpp
```diff
@@ -261,7 +261,7 @@ bool CallerIsInstrumented(void *sym) {
 static void OnStackUnwind(const SignalContext &sig, const void *,
                           BufferedStackTrace *stack) {
   stack->Unwind(StackTrace::GetNextInstructionPc(sig.pc), sig.bp, sig.context,
-                common_flags()->fast_unwind_on_fatal, GetStackTraceLen());
+                /*request_fast=*/true, GetStackTraceLen());
```
icmccorm, line 264: Is this because the slow unwinder is capturing parts of the sanitizer runtime?

icmccorm (approved):

bsan-rt/llvm-wrapper/bsan.cpp
vnt1c: Yes, it was capturing runtime frames from bsan's own code, and since I bounded the stack trace, it wasn't capturing the actual user program stack trace (maybe there is a better way to re-implement this?)

tests/ui.rs
vnt1c: Yes, you can take a look at the [reading_half_a_pointer.rs](https://github.com/BorrowSanitizer/bsan/blob/5e4962462f5011be9e5b508553f0d4c964e8d964/tests/miri-tests/should-fail/baseline/reading_half_a_pointer.rs#L26) test case for example, which constructs a non-canonical pointer, which causes different reporting on x86 (`x` in `reading_half_a_pointer` breaks the canonical rule and raises a general protection fault). Since ARM CPUs don't use the same canonical-address rule, they don't have the same issue.

PR #356, opened 2026-09-16
Track dead nodes with an is_dead flag

### LINKED ISSUE #340: Cache liveness within the state of each `Node`
The primary task of our GC is to create a "pending" set of nodes that are no longer reachable and can be pruned. If we cannot prune a node during a particular GC pass, then the pending set will retain it for subsequent collections. When we *can* prune a node, we will zero out its entry in the pending set, [like so](https://github.com/BorrowSanitizer/bsan/blob/1b65853587481dc52c715f64910fdf59e8cd6c60/bsan-rt/src/tree_borrows/tree.rs#L668-L669):
```rust
self.remove_useless_node(idx);
*entry = BorTag::omnivalid();
```
This ends up being somewhat clunky. Every time we finish pruning, we need to scan the entire contents of the pending set to identify which tags to keep, and which to throw away. It also wastes memory, since we are storing the entire value of the tag to indicate that a node is dead. This is a boolean state, and could be one bit.

Instead, we should add an `is_dead` flag to every `Node`. Every node identified as dead in the list of tags passed to `__bsan_prune` should have its flag set. This makes pruning much cleaner on the C++ side. It can be implemented as a single pass through the pending set:
```C++
pending_.drain([&](AllocInfo *info, BorTagSet &tags) {
  if (__bsan_prune(info, tags.data(), tags.size())) {
    quarantine_[info] = snap.gen;
  }
});
```
This makes pruning a little *more* complicated on the Rust side, though. It is no longer sufficient to prune *only* the nodes identified by the set of tags passed to `__bsan_prune`. These are only a subset of the nodes that might be prunable on a given pass. Consider that for a node to be pruned, it needs to satisfy two predicates:

1) It must be dead.

2) Its permission must be able to be replaced by the permissions of its children. 

The first predicate does not imply the second. Each time we call `__bsan_prune`, there may be nodes that were already marked as dead in a prior pass, and are absent from the list of tags in this pass, but are now prunable. We need to scan the tree to identify these nodes. Otherwise, we will leak them indefinitely.

vnt1c:
Closes #340.

Adds an `is_dead` flag to `Node`, so C++ no longer keeps unpruned tags in `pending_`. `__bsan_prune` marks the tags it receives as dead, then prunes every dead node in the tree from largest tag to smallest, including ones marked in earlier passes. C++ now drains `pending_` in a single pass. Updated the pruning tests and added tests for live tags not being marked dead and for compacting a dead node once a sibling is pruned.


bsan-rt/src/tree_borrows/tree.rs
```diff
@@ -172,6 +172,9 @@ pub struct Node {
     default_initial_idempotent_foreign_access: IdempotentForeignAccess,
     /// Whether a wildcard access could happen through this node.
     pub is_exposed: bool,
+    /// Whether this node is still reachable in memory. Note that nodes with a zero reference
```
icmccorm: When a node is dead, we know that it is actually not reachable from the shadow stack anymore, either!

mojeanmac:
In the extremely-long-runtime cases, there is definitely merit in this approach! I think we need to adapt this branch further to balance dead checks via tree traversal vs a deadlist.

I think the case is that the GC pass now takes longer, and thus it isn't advantageous to run it as often (which is negligible in high-runtime cases), so we could possibly change the interval to have it run less.

image
Per-crate seconds averaged over 120 crates:

| Config | Mean (s) | Geomean (s) |
| ------ | -------- | ----------- |
| bsan   | 1320.200 | 35.066      |
| branch | 1092.602 | 45.231      |

icmccorm:
Here's a possible fix: we prune leaves during GC pauses, and prune interior nodes during traversals.

When we prune a tree, we have a list of dead tags. We iterate over the list in reverse order (largest tag to smallest tag), and remove only the nodes that we can. Interior nodes that cannot be removed are marked dead. We do *not* traverse the tree to remove interior nodes that were marked as dead in prior passes. Then, in future traversals, if we visit an interior node that is marked dead, and can be removed, we prune it.

mojeanmac:
> Here's a possible fix: we prune leaves during GC pauses, and prune interior nodes during traversals.
> 
> When we prune a tree, we have a list of dead tags. We iterate over the list in reverse order (largest tag to smallest tag), and remove only the nodes that we can. Interior nodes that cannot be removed are marked dead. We do _not_ traverse the tree to remove interior nodes that were marked as dead in prior passes. Then, in future traversals, if we visit an interior node that is marked dead, and can be removed, we prune it.

I feel like persistently checking dead nodes that fail the permission relation check during regular traversal won't benefit runtime; we might be checking stubborn dead nodes more often than if it was just every GC pass. We'd also have to stop the world every time too.

This does have the merit of avoiding checking dead nodes on unused paths, though.

icmccorm:
> We might be checking stubborn dead nodes more often than if it was just every GC pass

We could tie this to the semantics of permission transitions. If a transition occurs that would unblock a dead parent, only then do we check.

> We'd also have to stop the world every time too.

We wouldn't need to stop the world in this case—once a node is marked dead via the flag, it's guaranteed to no longer be reachable anywhere, so we don't need to worry about destroying something that's been used somewhere else.

vnt1c:
> In the extremely-long-runtime cases, there is definitely merit in this approach! I think we need to adapt this branch further to balance dead checks via tree traversal vs a deadlist.

I agree with this! The data looks promising, at least based on the runtime of the larger crates, and I think a reasonable fix for the smaller crates may be to maintain a list for each tree's dead but unprunable tags. This way, we wouldn't need to re-scan the entire tree but rather pull from this list to perform the pruning checks.

We can replace the [full tree scan](https://github.com/BorrowSanitizer/bsan/pull/356/changes#diff-392dbe87566a4c5ff495e3a57294b1bb42b003e8e696444ce4210b1460cd2a67R645) with a simple iteration over the nodes in the `dead_but_unprunable` list and tags carried over from `dead_tags` and perform checks on those nodes. This brings the time down from O(N) to O(D).

